PR #40, opened 2015-02-04
Make JSON responses out of HttpExceptions (incl. headers)

rjkip:
a


DRvanR:
image

Nice Catch! :+1:

PR #45, opened 2015-03-02
Keep a searchable list of second factors for in the RA

rjkip:


rjkip:
:bell: @DRvanR

src/Surfnet/StepupMiddleware/ApiBundle/Controller/RaSecondFactorController.php
```diff
@@ -0,0 +1,59 @@
...
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpKernel\Exception\AccessDeniedHttpException;
+
+final class RaSecondFactorController extends Controller
+{
+    public function collectionAction(Request $request, Institution $institution)
+    {
+        if (!$this->isGranted('ROLE_RA')) {
+            throw new AccessDeniedHttpException('Client is not authorised to access identity');
```
DRvanR: Message could be updated, maybe even removed

src/Surfnet/StepupMiddleware/ApiBundle/Identity/Entity/RaSecondFactor.php
```diff
@@ -0,0 +1,143 @@
...
+ *          @ORM\Index(name="idx_ra_second_factor_email", columns={"email"}),
+ *      }
+ * )
+ */
+final class RaSecondFactor implements JsonSerializable
+{
+    const STATUS_UNVERIFIED = 'unverified';
+    const STATUS_VERIFIED = 'verified';
+    const STATUS_VETTED = 'vetted';
+    const STATUS_REVOKED = 'revoked';
```
DRvanR: I smell a value object :wink:

src/Surfnet/StepupMiddleware/ApiBundle/Identity/Entity/RaSecondFactor.php
```diff
@@ -0,0 +1,143 @@
...
+ * @ORM\Entity(repositoryClass="Surfnet\StepupMiddleware\ApiBundle\Identity\Repository\RaSecondFactorRepository")
+ * @ORM\Table(
+ *      indexes={
+ *          @ORM\Index(name="idx_ra_second_factor_type", columns={"type"}),
+ *          @ORM\Index(name="idx_ra_second_factor_second_factor_id", columns={"second_factor_id"}),
+ *          @ORM\Index(name="idx_ra_second_factor_status", columns={"status"}),
+ *          @ORM\Index(name="idx_ra_second_factor_identity_id", columns={"identity_id"}),
+ *          @ORM\Index(name="idx_ra_second_factor_institution", columns={"institution"}),
+ *          @ORM\Index(name="idx_ra_second_factor_name", columns={"name"}),
+ *          @ORM\Index(name="idx_ra_second_factor_email", columns={"email"}),
```
DRvanR: Not quite sure how I feel about the indexes... Esp. `status` and `type` might not benefit from an index due to lack of different values, but does incur the cost of having to maintain the index. Furthermore it seems that `name` and `email` would benefit from a `FULLTEXT` index. Also iirc not all these fields are searchable and thus would not need an index? Lastly, compound indexes covering the most common usecases would be very beneficial. Thoughts?

src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/RaSecondFactorRepository.php
```diff
@@ -0,0 +1,133 @@
...
+            case 'email':
+                $queryBuilder->orderBy("sf.$command->orderBy", $command->orderDirection === 'desc' ? 'DESC' : 'ASC');
+                break;
+            case 'status':
+                $queryBuilder
+                    ->addSelect('
+                        CASE WHEN sf.status=:unverified THEN 0 ELSE
+                        CASE WHEN sf.status=:verified THEN 1 ELSE
+                        CASE WHEN sf.status=:vetted THEN 2 ELSE
+                        CASE WHEN sf.status=:revoked THEN 3 ELSE 4
```
DRvanR: Iirc this should be written as
```sql
CASE sf.status
    WHEN :unverified THEN 0
    WHEN :verified THEN 1
    WHEN :vetted THEN 2
    WHEN :revoked THEN 3
    ELSE 4
END CASE
```
rather than nest case statements, or does the QueryBuilder not accept that?

DRvanR: Also, since it are static values (consts in a class), I'd not use parameters but set the values immediately.

DRvanR:
Nice work on the SecondFactorStatus.  Consider the typehint and please update or remove the message (see comment above). When that is done :+1:

PR #48, opened 2015-03-12
Expand columns for second factor identifier

DRvanR:


DRvanR:
@rjkip Not sure if you've got the same, but needed it myself anyways :smile:

rjkip:
Darn you! :+1:

PR #50, opened 2015-03-13
Ensure displayed ART and logged ART are equal.

rjkip:
This was fixed in the Stepup bundle, but not installed in all apps.


DRvanR:
:+1:

PR #61, opened 2015-03-24
Let authorities vet identities and only in case of sufficient LoA

rjkip:


rjkip:
:facepunch: @DRvanR

src/Surfnet/StepupMiddleware/CommandHandlingBundle/Tests/Identity/CommandHandler/IdentityCommandHandlerTest.php
```diff
@@ -631,4 +634,202 @@ public function an_unverified_second_factor_can_be_revoked()
...
+                    $authorityId,
+                    $authorityInstitution,
+                    $authorityPhoneSfId,
+                    DateTime::now(),
+                    'regcode',
+                    'Charlie Parker',
+                    'e@mail.com',
+                    'en_GB'
+                ),
+                new SecondFactorVettedEvent(
```
DRvanR: Seems simpler to use the bootstrap event?

rjkip: The bootstrap event bootstraps a Yubikey, which has a higher LoA.

src/Surfnet/Stepup/Identity/Identity.php
```diff
@@ -235,24 +238,74 @@ public function verifyEmail($verificationNonce)
...
-            )) {
-                continue;
-            }
+    public function vetSecondFactor(
+        IdentityApi $identity,
+        SecondFactorId $secondFactorId,
+        $secondFactorIdentifier,
+        $registrationCode,
+        $documentNumber,
+        $identityVerified
```
DRvanR: Would be nice to wrap the last three params in a VO rather then keep passing em around. Also, I wonder if `identityVerified` should be passed down all the way to the SF, rather than do it here. Semantically it might be right to do as well: `If I have not verified the Identity, I may not vet this SecondFactor`

src/Surfnet/Stepup/Identity/Identity.php
```diff
@@ -235,24 +238,74 @@ public function verifyEmail($verificationNonce)
...
+    ) {
+        /** @var VettedSecondFactor|false $secondFactorWithHighestLoa */
+        $secondFactorWithHighestLoa = array_reduce(
+            $this->vettedSecondFactors->toArray(),
+            function (VettedSecondFactor $carry, VettedSecondFactor $item) {
+                return $carry->hasEqualOrHigherLoaComparedTo($item) ? $carry : $item;
+            },
+            $this->vettedSecondFactors->first()
+        );
```
DRvanR: What would you think about moving this into a custom `SecondFactorCollection` that extends `ArrayCollection` or something. Seems it might be a bit cleaner? Fairly complex code to read... just a thought.

DRvanR: Also, would be easier to mask/abstract that this currently doesn't do anything, since there can be only one (Vetted)SecondFactor at this time....

src/Surfnet/Stepup/Identity/Identity.php
```diff
@@ -235,24 +238,74 @@ public function verifyEmail($verificationNonce)
...
 
-        throw new DomainException(
-            "Cannot vet second factor: no second factor with given registration code or registration window has closed."
+        $secondFactor->vet(
+            $registrationCode,
+            $secondFactorIdentifier,
+            $documentNumber,
+            $identityVerified,
+            $this->vettedSecondFactors,
+            $this->vettedSecondFactors
```
DRvanR: Can't seem to find the usage of these additional params?

src/Surfnet/Stepup/Identity/Identity.php
```diff
@@ -235,24 +238,74 @@ public function verifyEmail($verificationNonce)
...
+
+        if (!$secondFactor->wouldBeVettedBy(
+            $registrationCode,
+            $secondFactorIdentifier,
+            $documentNumber,
+            $identityVerified
+        )) {
+            throw new DomainException(
+                'Cannot vet second factor: registration code is incorrect or registration window has closed.'
+            );
```
DRvanR: Either the checks should be done here and the DomainException should be made precise, of this should be guards of invariants in the SecondFactor and the specific exceptions should be thrown there. Reading the previous sentence to me there is only one solution, kinda like with the email verification. Thoughts?

src/Surfnet/Stepup/Identity/Identity.php
```diff
@@ -235,24 +238,74 @@ public function verifyEmail($verificationNonce)
...
-                $secondFactorIdentifier,
-                $documentNumber,
-                $identityVerified
-            )) {
-                continue;
-            }
+    public function vetSecondFactor(
+        IdentityApi $identity,
+        SecondFactorId $secondFactorId,
+        $secondFactorIdentifier,
```
DRvanR: I would recommend renaming `$identity` to `$registrant` (indicative of the fact that that identity is the person standing at the desk requesting the vetting). Same with `$secondFactorId` and `$secondFactorType`, it should be clear that those are related to the registrant and not this identity :wink:

src/Surfnet/Stepup/Identity/Identity.php
```diff
@@ -235,24 +238,74 @@ public function verifyEmail($verificationNonce)
...
-            return;
+        $verifiedSecondFactorHasEqualOrLowerLoaComparedTo =
+            $identity->verifiedSecondFactorHasEqualOrLowerLoaComparedTo(
+                $secondFactorId,
+                $secondFactorWithHighestLoa
+            );
+
+        if (!$verifiedSecondFactorHasEqualOrLowerLoaComparedTo) {
+            throw new DomainException('Authority does not have the required LoA to vet the identity\'s second factor');
+        }
```
DRvanR: Wall of Text inc!

TL;DR: not sure how I feel about this all, but it's just a feeling. Have a different implementation below, how would you feel about something like that?

---

In all honesty, I'm not sure how I feel about these lines.... I've been hesitant about whether or not I should say something since it's just a feeling and it may well be that it's since I'm reading it on github, but these lines (and their consequences) are very hard to read/grok; amongst things due to the long variable and method names and the triple dispatch behind it. 
It kinda feels like that it could be simpler by just requesting the Verified Second Factor from the Identity and checking the LoA's of both... something like:

```php
$secondFactorToBeVetted = $identityToBeVetted->getVerifiedSecondFactor($sfid);
if (!$mySecondFactor->hasLoaHigherOrEqualTo($secondFactorToBeVetted) {
    throw new DomainException();
}
```
Which is somewhat analogous to what is happening in reality... 
The upside of something like that would also be that you get to call `$secondFactorToBeVetted->vet(/* params */)` albeit that might well be a bridge too far...

Also, `$identity->complyWithVettingOfSecondFactor` could also be `$identity->complyWithVetting()` as the Identity itself is vetted as well :wink:

---

With the renaming and SecondFactorCollection this would become something like:
```php
// when we have to check based on current LoA (future) this might become 
// $this->vettedSecondFactors->get('xxx');
$mySecondFactor = $this->vettedSecondFactors->getSecondFactorWithHighestLoa();
$registrantSecondFactor = $registrant->getVerifiedSecondFactor($registrantSecondFactorId);

if (!$mySecondFactor->hasLoaHigherThanOrEqualTo($registrantSecondFactor()) {
    throw new DomainException('naughty boy!');
}

$registrant->complyWithVetting(/* params */);
```

Thoughts?

src/Surfnet/Stepup/Identity/Identity.php
rjkip: If I put this in a VO, I'd not make things difficult by not passing it to the second factor.

src/Surfnet/StepupMiddleware/CommandHandlingBundle/Tests/Identity/CommandHandler/IdentityCommandHandlerTest.php
DRvanR: check

DRvanR: sucks when doing stuff good makes other things harder :tongue:

src/Surfnet/Stepup/Identity/Identity.php
DRvanR: It's not about difficulty, it's about whose responsibility it is... Even if it is part of the VO, should it be the responsibility of the second-factor-to-be-vetted to check whether or not the identity on which this method is invoked (authority) has actually verified the identity (registrant) that second-factor-to-be-vetted belongs to, hence the question (akin to the remarks about the `complyWithVetting` check/exception)

src/Surfnet/Stepup/Identity/Identity.php
rjkip: Moving into a custom collection sounds fine. What do you mean by "masking/abstracting that it currently doesn't do anything"?

src/Surfnet/Stepup/Identity/Identity.php
rjkip: The difference between e-mail verification and vetting is that e-mail verification may verify one of the potentially multiple second factors. In the case of vetting, the exact second factor is known beforehand, which eliminates the need to check from the outside whether the second factor would be vetted by. `VerifiedSecondFactor#vet()` checks the preconditions itself, throwing specific `DomainException`s. I posit there's actually no need for up-front checking of vettability and all these checks can be removed.

src/Surfnet/Stepup/Identity/Identity.php
rjkip: I assume you mean `$secondFactorIdentifier` instead of `$secondFactorType`?

src/Surfnet/Stepup/Identity/Identity.php
DRvanR: There is currently exactly one vetted second factor, rendering the `array_reduce` a bit redundant, not much to reduce if there is just one element - this could be masked by using the custom collection that actually returns just the one element it holds directly rather than using the reduce. Not that that has to happen, but would be fine to do.

src/Surfnet/Stepup/Identity/Identity.php
rjkip: I agree with fetching the verified second factor to remove the triple dispatch.
I disagree with calling `$sf->vet()`, let's not mess with child entities from the outside.
I disagree with renaming the method to `complyWithVetting` for consistency reasons: the caller is called `vetSecondFactor` and the event (important, hard to change) is called `SecondFactorVetted`.

DRvanR: Fair enough, good arguments :+1:

src/Surfnet/Stepup/Identity/Identity.php
DRvanR: yup, my bad

src/Surfnet/Stepup/Identity/Entity/SecondFactorCollection.php
```diff
@@ -0,0 +1,38 @@
...
+{
+    /**
+     * @return SecondFactor|null
+     */
+    public function getSecondFactorWithHighestLoa()
+    {
+        return array_reduce(
+            $this->toArray(),
+            function (SecondFactor $carry, SecondFactor $item) {
+                return $carry->hasEqualOrHigherLoaComparedTo($item) ? $carry : $item;
```
rjkip: I kept the reduction as the collection is unaware that it currently will contain 0..1 second factors.

rjkip:
Better? @DRvanR :dancer:

src/Surfnet/Stepup/Identity/Entity/SecondFactorCollection.php
DRvanR: fair enough

src/Surfnet/Stepup/Identity/Entity/SecondFactorCollection.php
```diff
@@ -0,0 +1,38 @@
...
+            },
+            $this->first() ?: null
```
DRvanR: Won't this break if it is `null`? (see typehint and method invocation in anonymous function)

src/Surfnet/Stepup/Identity/Entity/VerifiedSecondFactor.php
```diff
@@ -159,6 +144,16 @@ public function vet($registrationCode, $secondFactorIdentifier, $documentNumber,
         );
     }
 
+    public function hasEqualOrHigherLoaComparedTo(SecondFactor $comparable)
```
DRvanR: repeated three times in the exact same manner, smells like abstraction could be done?

src/Surfnet/Stepup/Identity/Identity.php
```diff
@@ -235,25 +235,64 @@ public function verifyEmail($verificationNonce)
...
+        $registrationCode,
+        $documentNumber,
+        $identityVerified
+    ) {
+        /** @var VettedSecondFactor|null $secondFactorWithHighestLoa */
+        $secondFactorWithHighestLoa = $this->vettedSecondFactors->getSecondFactorWithHighestLoa();
+        $registrantsSecondFactor = $registrant->getVerifiedSecondFactor($registrantsSecondFactorId);
+
+        if ($registrantsSecondFactor === null) {
+            throw new DomainException("Registrant second factor by given ID does not exist");
```
DRvanR: sprintf the ID in?

src/Surfnet/Stepup/Identity/Identity.php
```diff
@@ -235,25 +235,64 @@ public function verifyEmail($verificationNonce)
...
+        }
+
+        if (!$secondFactorWithHighestLoa->hasEqualOrHigherLoaComparedTo($registrantsSecondFactor)) {
+            throw new DomainException('Authority does not have the required LoA to vet the identity\'s second factor');
```
DRvanR: `s/identity's/registrants/`

src/Surfnet/StepupMiddleware/CommandHandlingBundle/Identity/CommandHandler/IdentityCommandHandler.php
```diff
@@ -175,15 +175,21 @@ public function handleVerifyEmailCommand(VerifyEmailCommand $command)
...
     {
+        /** @var IdentityApi $authority */
+        $authority = $this->repository->load(new IdentityId($command->authorityId));
         /** @var IdentityApi $identity */
         $identity = $this->repository->load(new IdentityId($command->identityId));
-        $identity->vetSecondFactor(
-            $command->registrationCode,
+
+        $authority->vetSecondFactor(
+            $identity,
```
DRvanR: registrant?

DRvanR:
Some minor details, looks good otherwise! :+1:

PR #76, opened 2015-05-18
Expanded AuditLog information

DRvanR:
Added the second factor identifier and the actor commonname.


DRvanR:
@rjkip lookie lookie

rjkip:
/giphy great
:+1:

PR #81, opened 2015-05-26
Allow an identity to express its locale preference

rjkip:
Requires DB zap

PR #82, opened 2015-05-29
Don't set any cookies to persist the user's security context

rjkip:
As the Middleware API is based on Basic authentication, no cookies
need to be set to keep the consumer authenticated. Also, no other
needs the be persisted from request to request.

Based on #81 so I don't have to keep wiping my event stream. Merge or cherry-pick into develop.


DRvanR:
:+1:

rjkip:
Merged into develop.

PR #83, opened 2015-06-01
Implement appoint as/retract RARole

DRvanR:


DRvanR:
:question:  @rjkip

rjkip:
I seems I just *don't have the power* to find anything to remark on.

(:+1:)

PR #92, opened 2015-06-17
Prevent shared reference between whitelist root and created event

rjkip:


rjkip:
:hamster: @DRvanR :hamster:

DRvanR:
:+1:

PR #94, opened 2015-06-18
Fix the check for institution for UpdateIdentityCommand

DRvanR:
depends on https://github.com/SURFnet/Stepup-Middleware-clientbundle/pull/32


rjkip:
Pending CI, :+1:

PR #112, opened 2016-06-03
Update README.md

relaxnow:
Clarify LOA resolution for Second Factor Only mode.

### DIFF
--- a/README.md
+++ b/README.md
@@ -309,9 +309,9 @@ Everything will be validated against the requirements listed above. Once the val
 
 ### LOA Resolution
 It is possible to specify a LOA in 3 places:
-1. The AuthnContextClassRef in the Authentication Request (SAML2 AuthnRequest).
-2. The ```loa``` on the ```service_providers``` configuration.
-3. The ```loa``` on the ```identity_providers``` configuration.
+  1. The AuthnContextClassRef in the Authentication Request (SAML2 AuthnRequest).
+  2. The ```loa``` on the ```service_providers``` configuration.
+  3. The ```loa``` on the ```identity_providers``` configuration.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -307,6 +307,15 @@ Each element in the service_providers array must be an object and contain the fo
...
 
+### LOA Resolution
+It is possible to specify a LOA in 3 places:
+1. The AuthnContextClassRef in the Authentication Request (SAML2 AuthnRequest).
+2. The ```loa``` on the ```service_providers``` configuration.
+3. The ```loa``` on the ```identity_providers``` configuration.
+
+The Gateway will authenticate the user with the highest LOA of all these.
+Second Factor Only mode requires that AuthnRequests use LOA aliases. However these are quickly translated to the equivalent of 'regular' LOAs. The configuration MUST NOT use Second Factor Only LOA aliases.
+
```
rjkip: Great that you guys are updating the documentation with things we forgot to document!

If you're documenting the LoA resolution "algorithm", I think you should go into the default LoAs per entity (`__default__`) and entity-specific LoAs (`https://entity.example/metadata`) as well.

> The Gateway will authenticate the user with the highest LOA of all these.
Maybe "The Gateway will require the user to use a second factor that provides an equal to or higher than the highest of these LoAs."?

Also: if you indent the numbered list, it is nicely formatted by GitHub :smile:

relaxnow: > Also: if you indent the numbered list, it is nicely formatted by GitHub 😄

Thanks for the tip!

relaxnow: As for the rest, I'm not particularly familiar with default LOA (only just had them explained to me) and entity specific LOAs. I'll come back to this if there is time at the end.

PR #116, opened 2016-06-09
Improve AuditLog API defaults and input validation

relaxnow:


rjkip:
:+1:

src/Surfnet/StepupMiddleware/ApiBundle/Controller/AuditLogController.php
```diff
@@ -31,11 +32,16 @@ public function secondFactorAuditLogAction(Request $request, Institution $instit
     {
         $this->denyAccessUnlessGranted(['ROLE_RA']);
 
+        $identityId = $request->get('identityId');
+        if (empty($identityId)) {
+            throw new BadApiRequestException(['This API-call MUST include the identityId as get parameter']);
```
DRvanR: Out of curiosity as to why this fix would be needed I wanted to check something for myself :wink: If I remember correctly, required parameters are (or rather, should) either be tackled on request parsing (paramconverter, as is done for `Institution`) or as a requirement in the URL (see various routes). It seems this was intended to be the latter:
the route [has the requirement][1], but that is not [correctly][2] implemented, and the parameter itself is missing, in the [bundle][3] as well... And it seems the identity [is indeed required][4].

Perhaps adapting the route to enforce it through an explicit routing parameter (`/audit-log/{identityId}/second-factors`) would be better in the long run? Thoughts @relaxnow @rjkip ?

(if so, a tech debt ticket would suffice i.m.o, especially since it is in internal API)

[1]: https://github.com/SURFnet/Stepup-Middleware/blob/16550f3852f98a9e7d36c5b3cfdc20d7565abc72/src/Surfnet/StepupMiddleware/ApiBundle/Resources/config/routing.yml#L95-L96
[2]: https://github.com/SURFnet/Stepup-Middleware/blob/16550f3852f98a9e7d36c5b3cfdc20d7565abc72/src/Surfnet/StepupMiddleware/ApiBundle/Resources/config/routing.yml#L79-L80
[3]: https://github.com/SURFnet/Stepup-Middleware-clientbundle/blob/6974c91e3e84b680f569680167e6b670a639ae54/src/Surfnet/StepupMiddlewareClient/Identity/Service/AuditLogService.php#L54
[4]: https://github.com/SURFnet/Stepup-Middleware/blob/16550f3852f98a9e7d36c5b3cfdc20d7565abc72/src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/AuditLogRepository.php#L65

rjkip: > **TL;DR** IMO, [update](https://github.com/SURFnet/Stepup-Middleware/blob/16550f3852f98a9e7d36c5b3cfdc20d7565abc72/src/Surfnet/StepupMiddleware/ApiBundle/Resources/config/routing.yml#L79-L80) route requirement to be a UUID, remove check in controller, and URL parameter would be nice, but register as tech debt.

I'm not sure whether Symfony enforces a requirement when the requirement is not a URL parameter, but rather a query parameter. We should verify that and [update](https://github.com/SURFnet/Stepup-Middleware/blob/16550f3852f98a9e7d36c5b3cfdc20d7565abc72/src/Surfnet/StepupMiddleware/ApiBundle/Resources/config/routing.yml#L79-L80) the requirement to a UUID. Since the audit log is per-identity, making the identity ID a URL parameter seems the logical choice, but then again, I remember us making conscious decision not to. However, I don't think we should go and change the route URL now, and tech debt would be fine.

Since we rely on the routing definition to impose certain requirements before routing the request to the controller action (for example, `Accept:application/json`), a query parameter requirement in the routing definition would suffice and an extra check would not be needed in the controller action.

PR #117, opened 2016-06-09
Always use UTC

relaxnow:
I'll probably need to take a look at the tests as they should all crash on any system that has a non-UTC tz set, but what do you think @rjkip ?

### DIFF
--- a/src/Surfnet/Stepup/DateTime/DateTime.php
+++ b/src/Surfnet/Stepup/DateTime/DateTime.php
@@ -20,6 +20,7 @@ namespace Surfnet\Stepup\DateTime;
 
 use DateInterval;
 use DateTime as CoreDateTime;
+use DateTimeZone;
 use Surfnet\Stepup\Exception\InvalidArgumentException;
 
 /**
@@ -50,7 +51,7 @@ class DateTime
      */
     public static function now()
     {
-        return self::$now ?: new self(new CoreDateTime);
+        return self::$now ?: new self(new CoreDateTime('now', new DateTimeZone('UTC')));
     }
 
     /**
@@ -63,7 +64,7 @@ class DateTime
             InvalidArgumentException::invalidType('string', 'dateTime', $string);
         }
 
-        $dateTime = CoreDateTime::createFromFormat(self::FORMAT, $string);
+        $dateTime = CoreDateTime::createFromFormat(self::FORMAT, $string, new DateTimeZone('UTC'));
 
         if ($dateTime === false) {
             throw new InvalidArgumentException('Date-time string could not be parsed: is it formatted correctly?');
@@ -77,7 +78,13 @@ class DateTime
      */
     public function __construct(CoreDateTime $dateTime = null)
     {
-        $this->dateTime = $dateTime ?: new CoreDateTime();
+        $dateTime = $dateTime ?: new CoreDateTime('now', new DateTimeZone('UTC'));
+
+        if ($dateTime->getOffset() !== 0) {
+            throw new InvalidArgumentException('Provided DateTime is not in UTC, but has offset: ' . $dateTime->getOffset());
+        }
+
+        $this->dateTime = $dateTime;
     }
 
     /**
--- a/src/Surfnet/Stepup/DateTime/DateTime.php
+++ b/src/Surfnet/Stepup/DateTime/DateTime.php
@@ -23,7 +23,6 @@ use DateTime as CoreDateTime;
 use Surfnet\Stepup\Exception\InvalidArgumentException;
 
 /**
- * @SuppressWarnings(PHPMD.TooManyMethods)
  * @SuppressWarnings(PHPMD.TooManyPublicMethods)
  */
 class DateTime
--- a/src/Surfnet/StepupMiddleware/ApiBundle/Identity/Projector/AuditLogProjector.php
+++ b/src/Surfnet/StepupMiddleware/ApiBundle/Identity/Projector/AuditLogProjector.php
@@ -100,12 +100,7 @@ class AuditLogProjector implements ProjectorInterface
         $entry->identityId          = (string) $auditLogMetadata->identityId;
         $entry->identityInstitution = $auditLogMetadata->identityInstitution;
         $entry->event               = get_class($event);
-        $entry->recordedOn          = new DateTime(
-            new CoreDateTime(
-                $domainMessage->getRecordedOn()->toString(),
-                new DateTimeZone('UTC')
-            )
-        );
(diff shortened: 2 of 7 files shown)


relaxnow:
@rjkip what do you think?

src/Surfnet/StepupMiddleware/ApiBundle/Tests/Identity/Projector/AuditLogProjectorTest.php
```diff
@@ -53,7 +54,7 @@ public function auditable_events()
                         new SecondFactorId('ijkl'),
                         new SecondFactorType('sms')
                     )),
-                    BroadwayDateTime::fromString('1970-01-01H00:00:00.000')
+                    BroadwayDateTime::fromString('1970-01-01T00:00:00.000')
```
rjkip: Lolwut, we must've been tired.

relaxnow: heh, yeah, the tests kept failing because the Stepup DateTime was given a DateTime with the "H" TimeZone. That kept me occupied for a while. 😛 
Stupid PHP for seeing that as a TZ.

src/Surfnet/StepupMiddleware/ApiBundle/Identity/Projector/AuditLogProjector.php
```diff
@@ -99,7 +100,12 @@ private function applyAuditableEvent(AuditableEvent $event, DomainMessage $domai
         $entry->identityId          = (string) $auditLogMetadata->identityId;
         $entry->identityInstitution = $auditLogMetadata->identityInstitution;
         $entry->event               = get_class($event);
-        $entry->recordedOn          = new DateTime(new CoreDateTime($domainMessage->getRecordedOn()->toString()));
+        $entry->recordedOn          = new DateTime(
```
rjkip: I'd like to understand why this was needed. I am under the impression Broadway does this properly.

src/Surfnet/Stepup/DateTime/DateTime.php
```diff
@@ -78,7 +78,13 @@ public static function fromString($string)
      */
     public function __construct(CoreDateTime $dateTime = null)
     {
-        $this->dateTime = $dateTime ?: new CoreDateTime();
+        $dateTime = $dateTime ?: new CoreDateTime('now', new DateTimeZone('UTC'));
+
+        if ($dateTime->getOffset() !== 0) {
+            throw new InvalidArgumentException('Provided DateTime is not in UTC, but has offset: ' . $dateTime->getOffset());
```
rjkip: This case isn't tested, or have I overlooked the test?

src/Surfnet/StepupMiddleware/ApiBundle/Identity/Projector/AuditLogProjector.php
relaxnow: Ah, just tested this again and it does. I was testing with the "H" formatted dates. Removed this change.

src/Surfnet/Stepup/DateTime/DateTime.php
relaxnow: Damnit, the dog ate my unit test commit, bad dog!
I (re-)added the test.

relaxnow:
@rjkip merge it? I'm a bit hesitant as I don't want to cause MW to come crashing in flames because somewhere unbeknownst to me non-UTC datetimes are still being used...

rjkip:
I'll check out this branch and see what it does to my set-up :wink:

rjkip:
@relaxnow Gateway won't run, since it requires PHP 5.6 and the box has PHP 5.4.

DRvanR:
Do note that this is a very dangerous fix, and potentially requires a lot more work to be done in either clients or in the command API. Any system that sends a command that has a DateTime (thankfully none right now iirc) either *must* send that as UTC or the command API *must* be able to detect datetimes and convert those to UTC if required before handling the command. Furthermore this assumes that all times currently are entered in UTC (which I believe is the case, but that is more by chance than intended :smile: )

If this fix (I'd rather call it a feature :wink:) is introduced, than the documentation **must** be adapted to outline this, and either the client or the middleware command api should be updated to at least accommodate the UTC conversion. Would hate see for instance the RA topple over because of a configuration difference in PHP timezones...

Out of curiosity: what problem does it try to solve?

relaxnow:
@DRvanR 
> Do note that this is a very dangerous fix, and potentially requires a lot more work to be done in either clients or in the command API. Any system that sends a command that has a DateTime (thankfully none right now iirc) either must send that as UTC or the command API must be able to detect datetimes and convert those to UTC if required before handling the command. 

Sure, sounds like worthwhile input validation for the MW to do.

> Furthermore this assumes that all times currently are entered in UTC (which I believe is the case, but that is more by chance than intended 😄 )

Do you know which times are entered?

> If this fix (I'd rather call it a feature 😉) is introduced, than the documentation must be adapted to outline this, and either the client or the middleware command api should be updated to at least accommodate the UTC conversion. Would hate see for instance the RA topple over because of a configuration difference in PHP timezones...

Sure.

> Out of curiosity: what problem does it try to solve?

Projections are stored in MySQL datetimes that lose TZ information so are read back in the server timezone which is CEST. Which sucks for 'recordedOn' which is created by Broadway in UTC.

Alternative fix might be that I'll change the DateTimeType from mysql datetime to string and store the TZ info with it. That might have significantly less impact.
What do you think @rjkip and @DRvanR ?

DRvanR:
> Projections are stored in MySQL datetimes that lose TZ information so are read back in the server timezone which is CEST. Which sucks for 'recordedOn' which is created by Broadway in UTC.

MariaDB stores datetimes without timezones, [unaffected by timezone][1] or conversions. Sadly, that indeed means we have to deal with PHP DateTime settings :cry: 

> Alternative fix might be that I'll change the DateTimeType from mysql datetime to string and store the TZ info with it. That might have significantly less impact.

yeah, but then you miss out on datetime selection criteria... Perhaps store it in an additional column (embeddable rather than type)? Not sure though...
In any case, thanks for filling me in, I'm just wary going down the road of enforcing timezones and having to deal with them explicitly :wink: 

[1]: https://mariadb.com/kb/en/mariadb/time-zones/#time-zone-effects

relaxnow:
Discussed this with @rjkip, instead of throwing an exception we'll convert the value to UTC and log a warning. We'll rename the class to make it explicit that we expect a UTC datetime and create a new issue to turn the warning into an exception after a couple of months of running in production.

PR #118, opened 2016-06-13
RA Locations: establishing the model

arothuis:
Part of [#116602125](https://www.pivotaltracker.com/story/show/116602125).

This PR is meant to get the model right before moving on to each separate Event, Command and the handling thereof.

### DIFF
--- a/src/Surfnet/Stepup/Configuration/Api/InstitutionConfiguration.php
+++ b/src/Surfnet/Stepup/Configuration/Api/InstitutionConfiguration.php
@@ -1,7 +1,7 @@
 <?php
 
 /**
- * Copyright 2015 SURFnet B.V.
+ * Copyright 2016 SURFnet B.V.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/src/Surfnet/Stepup/Configuration/Api/InstitutionConfiguration.php
+++ b/src/Surfnet/Stepup/Configuration/Api/InstitutionConfiguration.php
@@ -1,0 +1,69 @@
+<?php
+
+/**
+ * Copyright 2015 SURFnet B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Surfnet\Stepup\Configuration\Api;
+
+use Broadway\Domain\AggregateRoot;
+use Surfnet\Stepup\Configuration\Value\ContactInformation;
+use Surfnet\Stepup\Configuration\Value\Institution;
+use Surfnet\Stepup\Configuration\Value\Location;
+use Surfnet\Stepup\Configuration\Value\RaLocationId;
+use Surfnet\Stepup\Configuration\Value\RaLocationName;
+
+interface InstitutionConfiguration extends AggregateRoot
+{
+    /**
+     * @param Institution $institution
+     * @return InstitutionConfiguration
+     */
+    public static function createFor(Institution $institution);
+
+    /**
+     * @param RaLocationId $raLocationId
+     * @param RaLocationName $raLocationName
+     * @param Location $location
+     * @param ContactInformation $contactInformation
+     * @return void
+     */
+    public function addRaLocation(
+        RaLocationId $raLocationId,
+        RaLocationName $raLocationName,
+        Location $location,
+        ContactInformation $contactInformation
+    );
+
+    /**
+     * @param RaLocationId $raLocationId
+     * @param RaLocationName $raLocationName
+     * @param Location $location
+     * @param ContactInformation $contactInformation
+     * @return void
+     */
+    public function amendRaLocation(
+        RaLocationId $raLocationId,
+        RaLocationName $raLocationName,
+        Location $location,
+        ContactInformation $contactInformation
+    );
+
+    /**
+     * @param RaLocationId $raLocationId
+     * @return void
+     */
+    public function removeRaLocation(RaLocationId $raLocationId);
+}
--- a/src/Surfnet/Stepup/Configuration/Api/InstitutionConfiguration.php
+++ b/src/Surfnet/Stepup/Configuration/Api/InstitutionConfiguration.php
@@ -21,6 +21,7 @@ namespace Surfnet\Stepup\Configuration\Api;
 use Broadway\Domain\AggregateRoot;
 use Surfnet\Stepup\Configuration\Value\ContactInformation;
 use Surfnet\Stepup\Configuration\Value\Institution;
+use Surfnet\Stepup\Configuration\Value\InstitutionConfigurationId;
 use Surfnet\Stepup\Configuration\Value\Location;
 use Surfnet\Stepup\Configuration\Value\RaLocationId;
 use Surfnet\Stepup\Configuration\Value\RaLocationName;
@@ -28,10 +29,11 @@ use Surfnet\Stepup\Configuration\Value\RaLocationName;
 interface InstitutionConfiguration extends AggregateRoot
 {
     /**
+     * @param InstitutionConfigurationId $institutionConfigurationId
      * @param Institution $institution
      * @return InstitutionConfiguration
      */
-    public static function createFor(Institution $institution);
+    public static function create(InstitutionConfigurationId $institutionConfigurationId, Institution $institution);
 
     /**
      * @param RaLocationId $raLocationId
@@ -54,7 +56,7 @@ interface InstitutionConfiguration extends AggregateRoot
      * @param ContactInformation $contactInformation
      * @return void
      */
-    public function amendRaLocation(
+    public function changeRaLocation(
         RaLocationId $raLocationId,
         RaLocationName $raLocationName,
         Location $location,
--- a/src/Surfnet/Stepup/Configuration/Entity/RaLocation.php
+++ b/src/Surfnet/Stepup/Configuration/Entity/RaLocation.php
@@ -73,6 +73,10 @@ class RaLocation
         $this->contactInformation = $contactInformation;
     }
 
+    /**
+     * @param RaLocationId $raLocationId
+     * @return bool
+     */
     public function hasRaLocationId(RaLocationId $raLocationId)
     {
         return $this->raLocationId->equals($raLocationId);
--- a/src/Surfnet/Stepup/Configuration/Entity/RaLocation.php
+++ b/src/Surfnet/Stepup/Configuration/Entity/RaLocation.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace Surfnet\Stepup\Identity\Entity;
+namespace Surfnet\Stepup\Configuration\Entity;
 
 use Surfnet\Stepup\Configuration\Value\ContactInformation;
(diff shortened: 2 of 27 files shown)


src/Surfnet/Stepup/Configuration/Api/InstitutionConfiguration.php
```diff
@@ -0,0 +1,71 @@
+<?php
+
+/**
+ * Copyright 2015 SURFnet B.V.
```
rjkip: :goberserk:

src/Surfnet/Stepup/Configuration/Entity/RaLocation.php
```diff
@@ -0,0 +1,97 @@
+<?php
+
```
rjkip: :feelsgood:

src/Surfnet/Stepup/Configuration/Entity/RaLocation.php
```diff
@@ -0,0 +1,97 @@
...
+        Location $location,
+        ContactInformation $contactInformation
+    ) {
+        $this->raLocationId       = $raLocationId;
+        $this->locationName       = $locationName;
+        $this->location           = $location;
+        $this->contactInformation = $contactInformation;
+    }
+
+    public function hasRaLocationId(RaLocationId $raLocationId)
```
rjkip: El bloc du doc

src/Surfnet/Stepup/Configuration/Value/RaLocationList.php
```diff
@@ -0,0 +1,112 @@
...
+        foreach ($raLocations as $raLocation) {
+            $this->add($raLocation);
+        }
+    }
+
+    /**
+     * @param RaLocationId $raLocationId
+     * @return bool
+     */
+    public function hasWith(RaLocationId $raLocationId)
```
rjkip: `hasWith`, `removeWith`, meh. Not very common names and confusing to me at first. In collections, `has` is usually called `contains`, so maybe `containsWithId`. And I'd prefer `removeWithId`. Not sure though.

src/Surfnet/Stepup/Configuration/Value/RaLocationList.php
```diff
@@ -0,0 +1,112 @@
...
+    {
+        foreach ($this->raLocations as $raLocation) {
+            if ($raLocation->getRaLocationId()->equals($raLocationId)) {
```
rjkip: What about `getId` and `hasId`?

src/Surfnet/Stepup/Configuration/Value/RaLocationList.php
```diff
@@ -0,0 +1,112 @@
...
+            'Cannot remove RaLocation with id "%s" from RaLocationList: it is not present',
+            $raLocationId
+        ));
+    }
+
+    /**
+     * @param RaLocationId $raLocationId
+     * @return RaLocation
+     */
+    public function getBy(RaLocationId $raLocationId)
```
rjkip: I prefer including the indirect object: `getById`. But that's me.

arothuis:
More thoughts, @rjkip?

src/Surfnet/Stepup/Configuration/Value/InstitutionConfigurationId.php
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    public function jsonSerialize()
+    {
+        return (string) $this;
+    }
+
+    public function __toString()
+    {
+        return $this->institutionConfigurationId;
```
arothuis: Property `institutionConfigurationId` does not contain a string but a `Uuid` as its the result of `Uuid::uuid5()`.

rjkip:
Please create some test regarding InstConf creation. Also, please put the event serialisation through the same paces as those from Identity. I think the test class includes SmokeTest in its name. You'll find your first bugs then ;)

arothuis:
Will create some tests for the `InstitutionConfiguration` and the event serialisation, although I'm not sure what you mean by "same paces".

arothuis:
@rjkip

src/Surfnet/Stepup/Configuration/Value/ContactInformation.php
```diff
@@ -0,0 +1,61 @@
...
+     * @return bool
+     */
+    public function equals(ContactInformation $otherContactInformation)
+    {
+        return $this->contactInformation === $otherContactInformation->contactInformation;
+    }
+
+    public function jsonSerialize()
+    {
+        return (string) $this;
```
rjkip: The JSON and string representations are coincidentally the same; its JSON representation is not its string representation per se. So please explicitly return the contact information itself :)

src/Surfnet/Stepup/Configuration/Event/NewInstitutionConfigurationCreatedEvent.php
```diff
@@ -0,0 +1,58 @@
...
+        return new self(
+            new InstitutionConfigurationId($data['institution_configuration_id']),
+            new Institution($data['institution'])
+        );
+    }
+
+    public function serialize()
+    {
+        return [
+            'institution_configuration_id' => (string) $this->institutionConfigurationId,
```
rjkip: We did use string casting on ID value objects near the start of this project, but in time we migrated to explicitly extracting values from ID value objects. A value object may for instance be a compound value (IP address + subnet mask), and you would retrieve these values separately. You can see we've reflected this decision in [`IdentityId`](https://github.com/SURFnet/Stepup-Middleware/blob/b6516e39ec80f667f239d99bfa80b913cc1ccf1b/src/Surfnet/Stepup/Identity/Value/IdentityId.php#L46), for example.

Also, in my opinion `__toString()` is for creating a string representation of the object itself. So for a compound value like an IP address + subnet mask, the string representation would contain both values. Basically `__toString` is very useful for logging etc., but it's not a very explicit manner of communicating with this object.

So for the earlier reasons and consistency's sake, please extract the value explicitly.

src/Surfnet/Stepup/Tests/Configuration/Event/EventSerializationAndDeserializationTest.php
```diff
@@ -0,0 +1,150 @@
...
+            // Configuration
+            'NewConfigurationCreatedEvent' => [
+                new NewConfigurationCreatedEvent(
+                    Configuration::CONFIGURATION_ID
+                )
+            ],
+            'ConfigurationUpdatedEvent' => [
+                new ConfigurationUpdatedEvent(
+                    Configuration::CONFIGURATION_ID,
+                    ['configurationKey' => 'configurationValue']
```
rjkip: Awesome! Thanks for writing these for the Configuration domain as well :)

src/Surfnet/Stepup/Tests/Configuration/InstitutionConfigurationTest.php
```diff
@@ -0,0 +1,264 @@
...
+
+        $institutionConfiguration = InstitutionConfiguration::create($institutionConfigurationId, $institution);
+        $institutionConfiguration->addRaLocation(
+            $raLocationId,
+            $raLocationName,
+            $location,
+            $contactInformation
+        );
+
+        $actualRaLocations = $institutionConfiguration->getRaLocations();
```
rjkip: We are using CQRS, so why read from the write model? In my opinion, events are the way this object communicates, and nothing's wrong with testing which events it emits.

src/Surfnet/Stepup/Tests/Configuration/InstitutionConfigurationTest.php
```diff
@@ -0,0 +1,264 @@
...
+        $this->assertEquals($institutionConfigurationId, $institutionConfiguration->getInstitutionConfigurationId());
+        $this->assertEquals($institution, $institutionConfiguration->getInstitution());
+        $this->assertEquals($expectedRaLocations, $institutionConfiguration->getRaLocations());
+    }
+
+    /**
+     * @test
+     * @group aggregate
+     */
+    public function an_ra_location_can_be_added_to_an_institution_configuration()
```
rjkip: What about the unhappy path? ;)

src/Surfnet/Stepup/Tests/Configuration/Value/RaLocationListTest.php
```diff
@@ -0,0 +1,155 @@
+<?php
+
+/**
+ * Copyright 2015 SURFnet B.V.
```
rjkip: Ha! Probably not that important, but... 2016 :wink:

src/Surfnet/Stepup/Configuration/Value/ContactInformation.php
arothuis: Will fix. Keep in mind that `ContactInformation` is a direct copy of Identity's `ContactInformation`, so I might fix it there as well.

src/Surfnet/Stepup/Configuration/Event/NewInstitutionConfigurationCreatedEvent.php
arothuis: I favor explicitness over implicit conversion as well, but I did it this way to maintain consistency, looking at the `Identity` related events: the `IdentityId`'s are all casted to string in these events.

src/Surfnet/Stepup/Tests/Configuration/InstitutionConfigurationTest.php
arothuis: I agree that events are the way this object communicates and I started out that way. However, it is a messy approach as you would have to get the uncommitted events from the aggregate. For each `DomainMessage` in there, the payload should be as expected. I do see value in asserting the right (sequence of) event(s) is produced.

I also see value in testing the state change within the object. A certain state change is expected when a certain method is called. These kinds of expectations should be tested. Whether or not this is done via CQRS or otherwise, to me, seems of less importance as this kind of behaviour is part of the unit under test.

src/Surfnet/Stepup/Tests/Configuration/InstitutionConfigurationTest.php
arothuis: Sorry for being such an optimist 🍷 <-- half full.

src/Surfnet/Stepup/Tests/Configuration/InstitutionConfigurationTest.php
arothuis: I was wrong. I accidentally made my `apply*` methods public, leading me to think they were part of the (public) API of the Aggregate. These methods should *not* be public. 

Therefore, the reasons for me to explicitly test it are rendered void. I will remove these tests, but keep the events.

src/Surfnet/Stepup/Configuration/InstitutionConfiguration.php
```diff
@@ -0,0 +1,240 @@
...
+    {
+        return $this->institutionConfigurationId;
+    }
+
+    /**
+     * @return Institution
+     */
+    public function getInstitution()
+    {
+        return $this->institution;
```
rjkip: You be needing these? :eyes:

rjkip:
Pending build, :+1:

src/Surfnet/Stepup/Configuration/Entity/RaLocation.php
```diff
@@ -0,0 +1,140 @@
...
+{
+    /**
+     * @var RaLocationId
+     */
+    private $raLocationId;
+
+    /**
+     * @var RaLocationName
+     */
+    private $raLocationName;
```
relaxnow: @rjkip / @arothuis why is this ```raLocationId``` instead of ```id``` of the ```RaLocation``` object? Does this mean the RA will have to specify  the ```orderBy``` as ```ra_location_id``` instead of ```id```?

arothuis: I did it to make explicit what type of `id` is required, but I see other entities use short names as well. 

Good call, let's change this to `id` instead of `raLocationId` and `name` instead of `raLocationName` and change the getters. I'll make a PR.

PR #119, opened 2016-06-14
Warn on missing email template

relaxnow:
Fixes event replaying for missing document numbers.

### DIFF
--- a/src/Surfnet/StepupMiddleware/CommandHandlingBundle/Identity/Service/SecondFactorMailService.php
+++ b/src/Surfnet/StepupMiddleware/CommandHandlingBundle/Identity/Service/SecondFactorMailService.php
@@ -293,7 +293,9 @@ final class SecondFactorMailService
             $this->logger->warning(
                 sprintf(
                     'Skipping sending mail because template configuration for "%s" in locale "%s" or "%s" is missing',
-                    [$name, $locale, $fallbackLocale]
+                    $name, 
+                    $locale, 
+                    $fallbackLocale
                 )
             );
         }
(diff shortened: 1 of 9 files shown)


src/Surfnet/StepupMiddleware/CommandHandlingBundle/Identity/Service/SecondFactorMailService.php
```diff
@@ -91,6 +109,8 @@ public function __construct(
         $this->emailVerificationUrlTemplate = $emailVerificationUrlTemplate;
         $this->emailTemplateService = $emailTemplateService;
         $this->fallbackLocale = $fallbackLocale;
+        $this->logger = $logger;
+        $this->warnOnMissingTemplateConfiguration = $warnOnMissingTemplateConfiguration;
```
rjkip: Would be nice to check this is in fact a boolean (an exception would point to a bug elsewhere). :+1: otherwise

relaxnow: You're right, that is nice. Added. Thanks!

src/Surfnet/StepupMiddleware/CommandHandlingBundle/Identity/Service/SecondFactorMailService.php
```diff
@@ -244,4 +268,28 @@ public function sendVettedEmail(
...
+    {
+        $emailTemplate = $this->emailTemplateService->findByName($name, $locale, $this->fallbackLocale);
+
+        if ($emailTemplate) {
+            return $emailTemplate;
+        }
+
+        if ($this->warnOnMissingTemplateConfiguration) {
+            $this->logger->warning(
+                'Skipping sending mail because template configuration is missing',
```
rjkip: Oh, if I'm still on time, I think it would be very helpful when scanning the log messages that the email template's name is included in the log message. Sorry!

If you agree, change it, otherwise, merge away :ship:

relaxnow: just in time

relaxnow: @rjkip but it's included in the context, why would it also need to be in the message?

rjkip:
:+1:

PR #124, opened 2016-06-17
Remove the RaCandidates by using the EntityManager

relaxnow:
This works around a Doctrine bug where it sees an entity as managed because it happens to reuse an object hash from an earlier created and destroyed object.

### DIFF
--- a/src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/RaCandidateRepository.php
+++ b/src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/RaCandidateRepository.php
@@ -52,12 +52,12 @@ class RaCandidateRepository extends EntityRepository
     }
 
     /**
-     * @param string[] $sraaList
+     * @param string[] $nameIds
      * @return void
      */
-    public function removeByNameIds($sraaList)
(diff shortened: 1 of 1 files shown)


src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/RaCandidateRepository.php
```diff
@@ -41,19 +41,33 @@ public function save(RaCandidate $raCandidate)
...
 
+        $this->getEntityManager()->remove($raCandidate);
         $this->getEntityManager()->flush();
     }
 
     /**
+     * @param string[] $sraaList
+     * @return void
+     */
+    public function removeByNameIds($sraaList)
```
rjkip: While you're in here, could you fix the weird naming? `sraaList` → `nameIds`

arothuis:
Well done :+1:

src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/RaCandidateRepository.php
relaxnow: Sure, done.

PR #125, opened 2016-06-17
Don't add RaCandidate if identity is already an RA

relaxnow:
An RA may choose to couple a new SF, this may not lead to a new candidacy.

@pmeulen do you agree with this solution?

### DIFF
--- a/src/Surfnet/StepupMiddleware/ApiBundle/Identity/Projector/RaCandidateProjector.php
+++ b/src/Surfnet/StepupMiddleware/ApiBundle/Identity/Projector/RaCandidateProjector.php
@@ -30,6 +30,7 @@ use Surfnet\Stepup\Identity\Event\VettedSecondFactorRevokedEvent;
 use Surfnet\Stepup\Identity\Event\YubikeySecondFactorBootstrappedEvent;
 use Surfnet\StepupMiddleware\ApiBundle\Identity\Entity\RaCandidate;
 use Surfnet\StepupMiddleware\ApiBundle\Identity\Repository\RaCandidateRepository;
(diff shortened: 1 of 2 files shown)


relaxnow:
@pmeulen agreed

relaxnow:
@rjkip review plz?

src/Surfnet/StepupMiddleware/ApiBundle/Identity/Projector/RaCandidateProjector.php
```diff
@@ -49,6 +56,10 @@ public function __construct(RaCandidateRepository $raCandidateRepository)
      */
     public function applySecondFactorVettedEvent(SecondFactorVettedEvent $event)
     {
+        if ($this->raListingRepository->findByIdentityId($event->identityId)) {
+            return;
```
rjkip: Very unfortunate we have to introduce a time coupling with another read model, but I don't see another solution, currently. Perhaps it would be wise to start adding the identity's RA status to the events pertaining to RA candidate nomination so we can depend on that information, but I'll have to think about that.

src/Surfnet/StepupMiddleware/ApiBundle/Resources/config/projection.yml
```diff
@@ -18,6 +18,7 @@ services:
             - @surfnet_stepup_middleware_api.repository.verified_second_factor
             - @surfnet_stepup_middleware_api.repository.vetted_second_factor
             - @surfnet_stepup_middleware_api.repository.identity
+            - @doctrine.orm.middleware_entity_manager
```
rjkip: Added by accident?

rjkip:
One small note on the entity manager. After that, :+1::ship:

PR #126, opened 2016-06-17
Always use UTC for projection datetimes

relaxnow:


rjkip:
Great. I added some extra notes about events and projections to the ticket. https://www.pivotaltracker.com/projects/1163646/stories/121758429

:ship:

rjkip:
Cancel that. Please remove the `UnverifiedSecondFactor#verificationNonceValidUntil`, because this datetime will now be shifted 1 or 2 hours, and we don't use it at the moment.

https://github.com/SURFnet/Stepup-Middleware/blob/07dfd156450073fe5c1597f4740119235bd921f2/src/Surfnet/StepupMiddleware/ApiBundle/Identity/Entity/UnverifiedSecondFactor.php#L74

relaxnow:
Ah yes, forgot about that. Done.

PR #129, opened 2016-06-21
Correct faulty migration as a column has been removed since diffing

arothuis:
See title. 
General tip: pull develop often, run migrations when needed.


rjkip:
:+1:

PR #131, opened 2016-06-22
Institution can be either in the attributes or in the query

relaxnow:


src/Surfnet/StepupMiddleware/ApiBundle/Request/ConfigurationInstitutionParamConverter.php
```diff
@@ -26,21 +26,29 @@
 
 class ConfigurationInstitutionParamConverter implements ParamConverterInterface
 {
+    const INSTITUTION = 'institution';
```
arothuis: Good call!

src/Surfnet/StepupMiddleware/ApiBundle/Request/ConfigurationInstitutionParamConverter.php
```diff
@@ -26,21 +26,29 @@
...
+
     public function apply(Request $request, ParamConverter $configuration)
     {
-        $requestAttributes = $request->attributes;
-        $institution = $requestAttributes->get('institution', false);
+        $institution = $request->attributes->get(self::INSTITUTION, false);
+        $request->attributes->remove(self::INSTITUTION);
+
+        if (!$institution) {
```
arothuis: Below, we are using `$institution === false`. 

This is done to make the `false` check explicit as we are setting it explicitly as a default value if the `get()` method fails.

relaxnow: Good call!

PR #136, opened 2016-07-13
Add configured institution projection

arothuis:


arothuis:
A start: @DRvanR, thoughts?

DRvanR:
looking good :smile:

PR #137, opened 2016-07-13
Events creating institution configurations

arothuis:


src/Surfnet/StepupMiddleware/ApiBundle/Configuration/EventListener/IdentityEventListener.php
```diff
@@ -0,0 +1,79 @@
...
+    public function __construct(ConfiguredInstitutionRepository $configuredInstitutionRepository, Pipeline $pipeline)
+    {
+        $this->configuredInstitutionRepository = $configuredInstitutionRepository;
+        $this->pipeline                        = $pipeline;
+    }
+
+    /**
+     * @param DomainMessage $domainMessage
+     */
+    public function handle(DomainMessage $domainMessage)
```
arothuis: This is actually the same behaviour as a Projector, yet it would be inaccurate to call these EventListeners projectors as they don't project anything.

This behaviour could be abstracted away to an abstract `EventListener` class that gets extended by implementations. That can be done now or whenever there is another EventListener added.

DRvanR: good call, completely agree.

src/Surfnet/StepupMiddleware/ApiBundle/Configuration/EventListener/WhitelistEventListener.php
```diff
@@ -0,0 +1,117 @@
...
+    /**
+     * @param Institution $institution
+     */
+    private function createInstitutionConfigurationFor(Institution $institution)
+    {
+        $command              = new CreateInstitutionConfigurationCommand();
+        $command->UUID        = (string) Uuid::uuid4();
+        $command->institution = $institution;
+
+        $this->pipeline->process($command);
```
DRvanR: seeing this, might want to think about doing this in a separate object, something like `$institutionConfigurationFactory->createFor($institution)`.

Main reason is the coupling between EventListeners and the CommandHandlingPipeline, that just doesn't really feel right...

src/Surfnet/StepupMiddleware/ApiBundle/Tests/Configuration/EventListener/WhitelistEventListenerTest.php
```diff
@@ -0,0 +1,269 @@
...
+use Surfnet\StepupMiddleware\ApiBundle\Configuration\EventListener\WhitelistEventListener;
+use Surfnet\StepupMiddleware\ApiBundle\Configuration\Repository\ConfiguredInstitutionRepository;
+use Surfnet\StepupMiddleware\CommandHandlingBundle\Configuration\Command\CreateInstitutionConfigurationCommand;
+use Surfnet\StepupMiddleware\CommandHandlingBundle\Pipeline\Pipeline;
+
+class WhitelistEventListenerTest extends TestCase
+{
+    /**
+     * @test
+     * @group event-listener
```
DRvanR: might also want to add institutionconfiguration as group?

src/Surfnet/StepupMiddleware/ApiBundle/Tests/Configuration/EventListener/WhitelistEventListenerTest.php
```diff
@@ -0,0 +1,269 @@
...
+        $expectedCommand->institution = new ConfigurationInstitution($newInstitution);
+
+        $pipelineMock = Mockery::mock(Pipeline::class);
+        $pipelineMock->shouldReceive('process')
+            ->once()
+            ->andReturnUsing(function(CreateInstitutionConfigurationCommand $actualCommand) use ($expectedCommand) {
+                $this->assertEquals($expectedCommand->institution, $actualCommand->institution);
+            });
+
+
```
DRvanR: one newline too much :wink:

DRvanR:
Some nits, a suggestion and some food for thought. Do like the PR overall, but would like that you think about disconnecting the EventListener (reactive, read side-ish) from the pipeline (active, write-side) through an intermediate of sorts.

DRvanR:
As for the build, as far as I'm concerned the PHP5.4 failing is acceptable as we're working on moving to 5.6|7.0 support.

src/Surfnet/StepupMiddleware/ApiBundle/Configuration/EventListener/WhitelistEventListener.php
arothuis: I agree, we would like something intermediary as we're operating on the borders between Event and Command. An added benefit is that this object can be reused wherever we want InstitutionConfigurations to be created. 

I think calling it a factory would be a bit atypical, though, as we are doing more than creating InstitutionConfiguration objects, we are also persisting them. 

I'm not sure, but we could call it a dispatcher as it translates methods (`createFor`) to commands and dispatches them.

DRvanR: A dispatcher leaks implementation details, plus it dispatches one specific command in a specific way. 
A factory is indeed creational, it was just a quick example :wink: You might almost want a mediator of some sorts...
Anyway, a decoupling here would be good.

arothuis:
I'm not sure about having separate `EventListeners` named after the Events they listen to instead of why they're listening to them.

arothuis:
@DRvanR, updated the PR with my idea. Can always change if necessary (especially naming, class placement).

Regarding the build, shall I remove 5.4 and 5.5 and (disallow failures for 7.0)?

DRvanR:
I'm unsure whether or not I like the naming (it is not a service that creates InstitutionConfigurations, it is an interface), and I'm quite sure that I do not like the placement (it seems more infrastructural than something that is part of the domain - more on the level of CommandHandler, or even lower). Perhaps give that some thought.

Other than that, :+1: 

w.r.t the build, sure.

arothuis:
It is pretty difficult improving the naming and I would like to prevent appending `Interface`. Suggestions are welcome.

Regarding the placement, I am uncertain I understand what you have in mind. The interface does not seem infrastructural to me. It describes that a InstitutionConfigurations can be created for a certain institution. To me, this seems part of the InstitutionConfiguration domain.

How the creation is handled, the implementation, seems of a more infrastructural nature as it uses a CommandHandling pipeline to do so. In line with that, I think it would be better to move `ApiBundle\Configuration\Dispatcher` to `CommandHandlingBundle\Configuration\Dispatcher`.

arothuis:
We could also go another way and introduce an `InsitutionConfigurationProcessManager` (instead of separate `Dispatchers` and `EventListeners`) which listens to events relevant to the `InstitutionConfiguration` process and sends commands for relevant aggregates. 

Coupling between listening and sending would not be that bad. This kind of cross-boundary message routing is to be expected of a [process manager](https://msdn.microsoft.com/en-us/library/jj591569.aspx) where it was not expected of a typical `EventListener`.

arothuis:
@DRvanR, lots of changes

app/DoctrineMigrations/Version20160719090052.php
```diff
@@ -0,0 +1,94 @@
...
+            function (WhitelistEntry $whitelistEntry) {
+                return $whitelistEntry->institution;
+            },
+            $this->getWhitelistEntryRepository()->findAll()
+        );
+        $identities = array_map(
+            function (Identity $identity) {
+                return $identity->institution;
+            },
+            $this->getIdentityRepository()->findAll()
```
DRvanR: wait wut?

DRvanR: as said a few times: use InstitutionListings, they contain a unique list of all institutions based on Identities, see https://github.com/SURFnet/Stepup-Middleware/blob/develop/src/Surfnet/StepupMiddleware/ApiBundle/Identity/Projector/InstitutionListingProjector.php#L36

app/DoctrineMigrations/Version20160719090052.php
```diff
@@ -0,0 +1,94 @@
...
+            $pipeline->process($createInstitutionConfigurationCommand);
+        }
+    }
+
+    /**
+     * @param Schema $schema
+     */
+    public function down(Schema $schema)
+    {
+        // No down-migration needed as the structure has not changed.
```
DRvanR: please add a message here indicating that this was a data only migration.

src/Surfnet/StepupMiddleware/CommandHandlingBundle/Configuration/Dispatcher/CreateInstitutionConfigurationCommandDispatcher.php
```diff
@@ -0,0 +1,47 @@
...
+
+namespace Surfnet\StepupMiddleware\CommandHandlingBundle\Configuration\Dispatcher;
+
+use Rhumsaa\Uuid\Uuid;
+use Surfnet\Stepup\Configuration\Api\InstitutionConfigurationCreationService;
+use Surfnet\Stepup\Configuration\Value\Institution;
+use Surfnet\StepupMiddleware\CommandHandlingBundle\Configuration\Command\CreateInstitutionConfigurationCommand;
+use Surfnet\StepupMiddleware\CommandHandlingBundle\Pipeline\Pipeline;
+
+final class CreateInstitutionConfigurationCommandDispatcher implements InstitutionConfigurationCreationService
```
DRvanR: shouldn't this and the interface be removed?

src/Surfnet/StepupMiddleware/CommandHandlingBundle/Tests/Configuration/Processor/InstitutionConfigurationProcessorTest.php
```diff
@@ -0,0 +1,349 @@
...
+        $repositoryMock->shouldReceive('hasConfigurationFor')
+            ->andReturn(false);
+
+        $firstExpectedInstitution  = new ConfigurationInstitution($firstInstitutionName);
+        $secondExpectedInstitution = new ConfigurationInstitution($secondInstitutionName);
+
+        $pipelineMock = Mockery::mock(Pipeline::class);
+        $pipelineMock->shouldReceive('process')
+            ->once()
+            ->andReturnUsing(function (CreateInstitutionConfigurationCommand $command) use ($firstExpectedInstitution) {
```
DRvanR: this is not what this method is intended for (see [the docs][mock]). Might be better to create a custom matcher. For reference on how to do something like that see [this matcher][1].

[mock]: http://docs.mockery.io/en/latest/reference/expectations.html
[1]: https://github.com/OpenConext/OpenConext-engineblock/blob/5.x-dev/tests/OpenConext/Mockery/Matcher/ValueObjectEqualsMatcher.php

src/Surfnet/StepupMiddleware/CommandHandlingBundle/Configuration/Dispatcher/CreateInstitutionConfigurationCommandDispatcher.php
arothuis: Yes.

app/DoctrineMigrations/Version20160719090052.php
arothuis: I must have been confused about what you meant. Thanks for pointing it out.

src/Surfnet/StepupMiddleware/CommandHandlingBundle/Tests/Configuration/Processor/InstitutionConfigurationProcessorTest.php
arothuis: Oo, didn't know about `Matchers` that is pretty cool! Thanks.

app/DoctrineMigrations/Version20160719090052.php
DRvanR: FYI: I mean, you can write a message to output here (`$this->getOutput()->write()` or something like that... iirc)

arothuis:
@DRvanR

DRvanR:
:+1:

PR #138, opened 2016-07-14
Add marker interfaces to InstitutionConfiguration commands

arothuis:
This way, we can restrict which services are allowed to invoke which command.

### DIFF
--- a/src/Surfnet/StepupMiddleware/CommandHandlingBundle/Configuration/Command/CreateInstitutionConfigurationCommand.php
+++ b/src/Surfnet/StepupMiddleware/CommandHandlingBundle/Configuration/Command/CreateInstitutionConfigurationCommand.php
@@ -20,6 +20,7 @@ namespace Surfnet\StepupMiddleware\CommandHandlingBundle\Configuration\Command;
 
 use Surfnet\StepupMiddleware\CommandHandlingBundle\Command\AbstractCommand;
 use Surfnet\StepupMiddleware\CommandHandlingBundle\Command\ManagementExecutable;
+use Symfony\Component\Validator\Constraints as Assert;
 
 class CreateInstitutionConfigurationCommand extends AbstractCommand implements ManagementExecutable
 {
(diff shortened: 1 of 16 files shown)


arothuis:
@DRvanR

src/Surfnet/StepupMiddleware/CommandHandlingBundle/Configuration/Command/CreateInstitutionConfigurationCommand.php
```diff
@@ -19,8 +19,9 @@
 namespace Surfnet\StepupMiddleware\CommandHandlingBundle\Configuration\Command;
 
 use Surfnet\StepupMiddleware\CommandHandlingBundle\Command\AbstractCommand;
+use Surfnet\StepupMiddleware\CommandHandlingBundle\Command\ManagementExecutable;
 
-class CreateInstitutionConfigurationCommand extends AbstractCommand
+class CreateInstitutionConfigurationCommand extends AbstractCommand implements ManagementExecutable
```
DRvanR: is that so? What is the current user if a new Identity leads to the discovery of a new Institution?

arothuis: I see what you mean. 

In our current application, `CreateInstitutionConfigurationCommands` are sent based on events caused by earlier commands. These earlier commands can be sent by any user (i.e. SelfService for `CreateIdentityCommand` and Middleware for `AddToWhitelistCommand`).

In the code, I assume the command ought to be created by Middleware since the incoming events and commands are handled there. I would have to check, but my assumption could be wrong because the earlier and resulting commands all happen within the same transaction, thus would bound to the same user.

DRvanR: Think theory vs practice: in reality, which user of the middleware will be the instigator of the creation of an InstitutionConfiguration, even if the creation is derived from other intent? The command should be executable on behalf of those users, to give some limitation.

arothuis:
I see my first merge got messed up by undo'ing it and re-applying it afterwards. There's more harm in rebasing than just leaving it in there, I guess.

arothuis:
@DRvanR

DRvanR:
:+1:

PR #139, opened 2016-07-20
Fix vetting locations and email templates

arothuis:
Note that the concept of a VettingLocation will have to go in the near future, as RaLocations and RegistrationAuthorityCredentials are clearly distinguishable concepts that do not need to be unified in this manner.


DRvanR:
:+1:

PR #140, opened 2016-07-20
Expand tables for event replays

arothuis:
I found out my event replays were consistently failing, because projection tables were not cleared automatically. This adds the tables to the EventStreamReplayer, causing them to be cleared properly before replaying.


arothuis:
@DRvanR

DRvanR:
:+1:

PR #142, opened 2016-07-22
Institution configuration options in aggregate

arothuis:


src/Surfnet/Stepup/Tests/Configuration/InstitutionConfigurationTest.php
```diff
@@ -0,0 +1,224 @@
...
+use Surfnet\Stepup\Configuration\Event\NewInstitutionConfigurationCreatedEvent;
+use Surfnet\Stepup\Configuration\Event\ShowRaaContactInformationOptionChangedEvent;
+use Surfnet\Stepup\Configuration\Event\UseRaLocationsOptionChangedEvent;
+use Surfnet\Stepup\Configuration\InstitutionConfiguration;
+use Surfnet\Stepup\Configuration\Value\Institution;
+use Surfnet\Stepup\Configuration\Value\InstitutionConfigurationId;
+use Surfnet\Stepup\Configuration\Value\ShowRaaContactInformationOption;
+use Surfnet\Stepup\Configuration\Value\UseRaLocationsOption;
+
+class InstitutionConfigurationTest extends AggregateRootScenarioTestCase
```
arothuis: These tests can move to a CommandHandler test once there is one.

arothuis:
@DRvanR 👀

src/Surfnet/Stepup/Configuration/Api/InstitutionConfiguration.php
```diff
@@ -25,15 +25,36 @@
...
+     * @param UseRaLocationsOption $useRaLocationsOption
+     * @return void
+     */
+    public function changeUseRaLocationsOption(UseRaLocationsOption $useRaLocationsOption);
+
+    /**
+     * @param ShowRaaContactInformationOption $raaContactInformationOption
+     * @return void
+     */
+    public function changeShowRaaContactInformationOption(ShowRaaContactInformationOption $raaContactInformationOption);
```
DRvanR: `set` vs `change`? Seems change would imply a change, regardless of whether or not there actually is a change... set is a bit of a nono, however, you're setting an option, something that is actually quite normal to do. Anyways, might want to give it some thought. Not saying it has to change, just that there might be a better alternative...

src/Surfnet/Stepup/Configuration/InstitutionConfiguration.php
```diff
@@ -58,15 +62,44 @@ class InstitutionConfiguration extends EventSourcedAggregateRoot implements Inst
...
      */
-    public static function create(InstitutionConfigurationId $institutionConfigurationId, Institution $institution)
-    {
+    public static function create(
+        InstitutionConfigurationId $institutionConfigurationId,
+        Institution $institution,
+        UseRaLocationsOption $useRaLocationsOption = null,
+        ShowRaaContactInformationOption $showRaaContactInformationOption = null
+    ) {
+        if ($useRaLocationsOption === null) {
```
DRvanR: why the nullable? Is there a way to create this *with* options?

DRvanR:
two lil' things.

arothuis:
@DRvanR, I removed the optional arguments.

You described my thought process on the `set` vs `change` exactly. In parlance, I would prefer setting an option. However, that could be confused with setters in code (which they explicitly are not). I also considered `toggle` but that feels more like the general process of inverting a binary option. For lack of a better alternative, I would keep it this way.

DRvanR:
:+1:

PR #143, opened 2016-07-22
Institution configuration option commands

arothuis:


arothuis:
@DRvanR

arothuis:
@DRvanR

DRvanR:
:+1:

DRvanR:
perhaps `Reconfigure` over `configure`? dunno...

arothuis:
I guess reconfigure makes sense, considering they are already configured by default. 
I'll rename 'em.

PR #144, opened 2016-07-26
Institution configuration options api

arothuis:
This adds an API for configuring options regarding institution configurations.
This also adds a JsonHelper class to reduce dependency on Guzzle.

The API for getting configurations from Middleware will be made once the read-side of institution configuration options is done.


src/Surfnet/StepupMiddleware/ManagementBundle/Tests/Controller/ConfigurationControllerTest.php
```diff
@@ -38,7 +38,7 @@ public function setUp()
         $this->loadFixtures([]);
 
         $this->client = static::createClient();
-        $this->password = $this->client->getKernel()->getContainer()->getParameter('management_password');
+        $this->password = 'management_password_test';
```
arothuis: The ConfigurationController currently does not test whether a 400 is returned when a command is invalid.

In fact, a 500 is thrown. The `InvalidCommandExceptionListener` is disregarded because the generic  `ExceptionListener` seems to have precedence over it. 

We might want to consider converting `InvalidCommandExceptions` to `BadCommandRequestExceptions` by hand in relevant controllers. This is out of scope for this PR.

arothuis:
@DRvanR

app/config/config_test.yml
```diff
@@ -1,5 +1,6 @@
 imports:
     - { resource: config_dev.yml }
+    - { resource: security_test.yml }
```
DRvanR: not saying it should change, but wouldn't configured known values be enough (say in `parameters.yml`)?

DRvanR: what I mean: this runs the risk of deviating from the prod version, with all possible ramifications

src/Surfnet/StepupMiddleware/ManagementBundle/Controller/InstitutionConfigurationController.php
```diff
@@ -0,0 +1,128 @@
...
+/**
+ * @SuppressWarnings(PHPMD.CouplingBetweenObjects)
+ */
+final class InstitutionConfigurationController extends Controller
+{
+    public function reconfigureAction(Request $request)
+    {
+        $this->denyAccessUnlessGranted(['ROLE_MANAGEMENT']);
+
+        $configuration = GuzzleHttp\json_decode($request->getContent(), true);
```
DRvanR: Might want to wrap this in a `JsonRequestHelper::decode` or something to ease the pain when migrating to a newer guzzle version

src/Surfnet/StepupMiddleware/ManagementBundle/Controller/InstitutionConfigurationController.php
```diff
@@ -0,0 +1,128 @@
...
+            $command                                  = new ReconfigureInstitutionConfigurationOptionsCommand();
+            $command->UUID                            = (string) Uuid::uuid4();
+            $command->institution                     = $institution;
+            $command->useRaLocationsOption            = $options['use_ra_locations'];
+            $command->showRaaContactInformationOption = $options['show_raa_contact_information'];
+
+            $commands[] = $command;
+        }
+
+        if (empty($commands)) {
```
DRvanR: This check could be done earlier, and then exit immediately as well (if it is relevant). Would suggest just to log the amount of commands that are going to be executed, not only log when no commands are going to be executed.

src/Surfnet/StepupMiddleware/ManagementBundle/Controller/InstitutionConfigurationController.php
```diff
@@ -0,0 +1,128 @@
...
+            'applied_at'   => (new DateTime())->format(DateTime::ISO8601),
+        ]);
+
+        return $response;
+    }
+
+    /**
+     * @param Pipeline $pipeline
+     * @param Command $command
+     * @return JsonResponse
```
DRvanR: orly? :wink:

src/Surfnet/StepupMiddleware/ManagementBundle/Controller/InstitutionConfigurationController.php
```diff
@@ -0,0 +1,128 @@
...
+            );
+        }
+    }
+
+    /**
+     * @return DataCollectingValidator
+     */
+    private function getValidator()
+    {
+        return $this->container->get('validator');
```
DRvanR: `->container` should not be necessary

src/Surfnet/StepupMiddleware/ManagementBundle/Controller/InstitutionConfigurationController.php
```diff
@@ -0,0 +1,128 @@
...
+    /**
+     * @return Logger
+     */
+    private function getLogger()
+    {
+        return $this->get('logger');
+    }
+
+    /**
+     * @return TransactionAwarePipeline
```
DRvanR: Please use the interface over an inmplementation

src/Surfnet/StepupMiddleware/ManagementBundle/Tests/Controller/ConfigurationControllerTest.php
DRvanR: the more I see this, the less I want this, please use the `parameters.yml` file. If needed, provide a fixture for test purposes instead. This reeks like magic strings that are related to something undiscoverable.

src/Surfnet/StepupMiddleware/ManagementBundle/Tests/Controller/InstitutionConfigurationControllerTest.php
```diff
@@ -0,0 +1,152 @@
...
+     */
+    private $password;
+
+    public function setUp()
+    {
+        // Initialises schema.
+        $this->loadFixtures([]);
+
+        $this->client = static::createClient();
+        $this->password = 'management_password_test';
```
DRvanR: same here, use the param over the hardcoded string.

src/Surfnet/StepupMiddleware/ManagementBundle/Tests/Controller/ConfigurationControllerTest.php
DRvanR: Please fix the issue at hand under boyscouting, please ensure the `InvalidCommandExecptionListener` is prioritized over the generic `ExceptionListener` (see prioroties listed [here][1])

[1]: http://symfony.com/doc/2.7/reference/events.html#kernel-exception

src/Surfnet/StepupMiddleware/ManagementBundle/Controller/InstitutionConfigurationController.php
arothuis: I'm all for wrapping it in a helper (our own implementation + Exception handling of this json_decode; I'm pretty sure we've already done this somewhere). 
I could change other occurences of this under boyscouting as well.

DRvanR: :+1:

src/Surfnet/StepupMiddleware/ManagementBundle/Controller/InstitutionConfigurationController.php
```diff
@@ -0,0 +1,152 @@
...
+                $connectionHelper->rollBack();
+
+                throw new AccessDeniedHttpException(
+                    sprintf('Processing of command "%s" is forbidden for this client', $command),
+                    $e
+                );
+            } catch (Exception $exception) {
+                $connectionHelper->rollBack();
+
+                throw $exception;
```
arothuis: Not sure about this. We're just intervening and throwing the exception out again, but maybe we can convert it to something more specific? Then again, we have an ExceptionListener that partially takes care of this.

arothuis:
@DRvanR, I changed it once more.

src/Surfnet/StepupMiddleware/ManagementBundle/Controller/InstitutionConfigurationController.php
DRvanR: What would you want to convert it to? We're just intercepting for the rollback, not because we actually want to reason about the exception :smile:

DRvanR:
Lookin' real good :+1: - nice work on the Jsonhelper, that ought to alleviate at least part of the guzzle migration headaches

src/Surfnet/StepupMiddleware/ManagementBundle/Controller/InstitutionConfigurationController.php
arothuis: Yea, that's true. My 'not sure about this' was more about the way this looks kind of funny (next to the more specific Exception conversion) than anything else. It does make sense, which is why I committed it, and the fact remains that we're merely intercepting in order to rollback.

PR #146, opened 2016-07-28
Show institution configuration options

arothuis:
* Corrected boolean type conversion in custom Doctrine types
* Corrected exception messages for type conversion
* Corrected quoting in `ApiBundle/Resources/config/services.yml`
* Added action to show the options for all InstitutionConfigurations:
```
GET management/institution-configuration

[
  {
    "dev.organisation.example": {
      "use_ra_locations": false,
      "show_raa_contact_information": true
    }
  },
 ...
  {
    "yubi.organisation.example": {
      "use_ra_locations": false,
      "show_raa_contact_information": true
    }
  }
]
```

### DIFF
--- a/src/Surfnet/StepupMiddleware/ApiBundle/Doctrine/Type/UseRaLocationsOptionType.php
+++ b/src/Surfnet/StepupMiddleware/ApiBundle/Doctrine/Type/UseRaLocationsOptionType.php
@@ -61,8 +61,15 @@ class UseRaLocationsOptionType extends Type
             return $value;
         }
 
+        // Prevent a too lenient boolean conversion on non-postgresql platforms
+        if ($platform->getName() !== 'postgresql') {
+            if ($value !== 1 && $value !== 0 && $value !== '1' && $value !== '0' && !is_bool($value)) {
+                throw ConversionException::conversionFailed($value, $this->getName());
+            }
+        }
+
         try {
-            $useRaLocationsOption = new UseRaLocationsOption($value);
+            $useRaLocationsOption = new UseRaLocationsOption($platform->convertFromBoolean($value));
         } catch (InvalidArgumentException $e) {
             // get nice standard message, so we can throw it keeping the exception chain
             $doctrineExceptionMessage = ConversionException::conversionFailedFormat(
--- a/src/Surfnet/StepupMiddleware/ApiBundle/Doctrine/Type/UseRaLocationsOptionType.php
+++ b/src/Surfnet/StepupMiddleware/ApiBundle/Doctrine/Type/UseRaLocationsOptionType.php
@@ -61,13 +61,6 @@ class UseRaLocationsOptionType extends Type
             return $value;
         }
 
-        // Prevent a too lenient boolean conversion on non-postgresql platforms
-        if ($platform->getName() !== 'postgresql') {
-            if ($value !== 1 && $value !== 0 && $value !== '1' && $value !== '0' && !is_bool($value)) {
-                throw ConversionException::conversionFailed($value, $this->getName());
-            }
-        }
-
         try {
             $useRaLocationsOption = new UseRaLocationsOption($platform->convertFromBoolean($value));
         } catch (InvalidArgumentException $e) {
--- a/src/Surfnet/StepupMiddleware/ApiBundle/Doctrine/Type/UseRaLocationsOptionType.php
+++ b/src/Surfnet/StepupMiddleware/ApiBundle/Doctrine/Type/UseRaLocationsOptionType.php
@@ -72,10 +72,9 @@ class UseRaLocationsOptionType extends Type
             $useRaLocationsOption = new UseRaLocationsOption($platform->convertFromBoolean($value));
         } catch (InvalidArgumentException $e) {
             // get nice standard message, so we can throw it keeping the exception chain
-            $doctrineExceptionMessage = ConversionException::conversionFailedFormat(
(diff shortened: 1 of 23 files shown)


src/Surfnet/StepupMiddleware/ApiBundle/Doctrine/Type/UseRaLocationsOptionType.php
```diff
@@ -61,14 +61,20 @@ public function convertToPHPValue($value, AbstractPlatform $platform)
             return $value;
         }
 
+        // Prevent a too lenient boolean conversion on non-postgresql platforms
+        if ($platform->getName() !== 'postgresql') {
+            if ($value !== 1 && $value !== 0 && $value !== '1' && $value !== '0' && !is_bool($value)) {
```
arothuis: It feels a bit like overkill doing this, but it also feels weird when a value of `string("hello world")` could theoretically be converted to `boolean(true)`.

arothuis:
@DRvanR, some additions, some fixes, some boyscouting

src/Surfnet/StepupMiddleware/ApiBundle/Doctrine/Type/UseRaLocationsOptionType.php
DRvanR: could that string be converted? I mean, how would it get into the database? Oh, and what about [oracle][1]? KISS - we do not support anything but MySQL (MariaDB).

[1]: https://docs.oracle.com/cd/B19306_01/olap.102/b14346/dml_datatypes004.htm#CJACGBGE

src/Surfnet/StepupMiddleware/ManagementBundle/Controller/InstitutionConfigurationController.php
```diff
@@ -42,6 +44,25 @@
  */
 final class InstitutionConfigurationController extends Controller
 {
+    public function showAction()
+    {
+        $this->denyAccessUnlessGranted(['ROLE_MANAGEMENT']);
+
+        $response = array_map(
+            function (InstitutionConfigurationOptions $options) {
```
DRvanR: how about implementing [this][1] (as is done in [about all others][2] :stuck_out_tongue_winking_eye: )

[1]: http://php.net/manual/en/class.jsonserializable.php
[2]: https://github.com/SURFnet/Stepup-Middleware/tree/develop/src/Surfnet/StepupMiddleware/ApiBundle/Identity/Entity

src/Surfnet/StepupMiddleware/ManagementBundle/Tests/Controller/InstitutionConfigurationControllerTest.php
```diff
@@ -66,6 +66,27 @@ public function authorization_is_required()
      * @test
      * @group management
      */
+    public function authorization_is_required_for_getting()
```
DRvanR: `getting`->`querying_of_institution_configuration_options` `- Describe what the intended behaviour is, not the practical implementation (same with `posting` above).

src/Surfnet/StepupMiddleware/ApiBundle/Doctrine/Type/UseRaLocationsOptionType.php
DRvanR: See https://github.com/SURFnet/Stepup-Middleware/blob/develop/app/DoctrineMigrations/Version20141119093423.php#L16 for instance

src/Surfnet/StepupMiddleware/ManagementBundle/Controller/InstitutionConfigurationController.php
arothuis: Sounds like a plan.

src/Surfnet/StepupMiddleware/ApiBundle/Doctrine/Type/UseRaLocationsOptionType.php
arothuis: Alright, I think `$platform->getFromBoolean($value)` would suffice. Given that we only speak to the database through the DBAL, this should not give any issues.

arothuis: I have removed the strict checking and embraced conversion through DBAL.

As an aside, Doctrine [converts booleans to INT(1) for Oracle][1]. It's kind of funny as they [should only fallback][2] to something like that if booleans were not natively supported. But maybe it is because they want to "[ensure the least possible data storage requirements are met][2]" regardless of native support.

The reason I chose to exclude PostgreSQL specifically was that Doctrine's `$platform` only has a custom `convertFromBoolean` method if it is an instance of `PostgreSqlPlatform`, which IIRC deals with a list of TRUE/FALSE representations of strings. In all other scenarios, `$platform->convertFromBoolean($value)` would equal `return null === $item ? null: (bool) $item ;`. 

[1]: http://doctrine-orm.readthedocs.io/projects/doctrine-dbal/en/latest/reference/types.html#mapping-matrix
[2]: http://doctrine-orm.readthedocs.io/projects/doctrine-dbal/en/latest/reference/types.html#boolean

arothuis:
@DRvanR

Turns out the entity was already JsonSerializable, just not in the way we actually want to represent it in JSON-form ;)

DRvanR:
:+1:

PR #147, opened 2016-07-28
Add API for institution configuration options for institution

arothuis:
This depends on the service introduced in #146 

```
GET /institution-configuration-options/dev.organisation.example

{
  "use_ra_locations": false,
  "show_raa_contact_information": true
}
```

### DIFF
--- a/src/Surfnet/StepupMiddleware/ApiBundle/Controller/InstitutionConfigurationOptionsController.php
+++ b/src/Surfnet/StepupMiddleware/ApiBundle/Controller/InstitutionConfigurationOptionsController.php
@@ -1,0 +1,52 @@
+<?php
+
+/**
+ * Copyright 2016 SURFnet B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Surfnet\StepupMiddleware\ApiBundle\Controller;
+
+use Surfnet\Stepup\Configuration\Value\Institution;
+use Surfnet\StepupMiddleware\ApiBundle\Service\InstitutionConfigurationOptionsService;
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Component\HttpFoundation\JsonResponse;
+
+final class InstitutionConfigurationOptionsController extends Controller
+{
+    public function getForInstitutionAction($institution)
+    {
+        $this->denyAccessUnlessGranted(['ROLE_SS', 'ROLE_RA']);
+
+        $institutionConfigurationOptions = $this
+            ->getInstitutionConfigurationOptionsService()
+            ->findInstitutionConfigurationOptionsFor(new Institution($institution));
+
+        return new JsonResponse(
+            [
+                'use_ra_locations'             => $institutionConfigurationOptions->useRaLocationsOption->isEnabled(),
+                'show_raa_contact_information' => $institutionConfigurationOptions
+                    ->showRaaContactInformationOption->isEnabled(),
+            ]
+        );
+    }
+
+    /**
+     * @return InstitutionConfigurationOptionsService
+     */
+    private function getInstitutionConfigurationOptionsService()
+    {
+        return $this->get('surfnet_stepup_middleware_api.service.institution_configuration_options');
+    }
+}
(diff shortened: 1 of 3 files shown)


src/Surfnet/StepupMiddleware/ApiBundle/Controller/InstitutionConfigurationOptionsController.php
```diff
@@ -0,0 +1,52 @@
...
+    {
+        $this->denyAccessUnlessGranted(['ROLE_SS', 'ROLE_RA']);
+
+        $institutionConfigurationOptions = $this
+            ->getInstitutionConfigurationOptionsService()
+            ->findInstitutionConfigurationOptionsFor(new Institution($institution));
+
+        return new JsonResponse(
+            [
+                'use_ra_locations'             => $institutionConfigurationOptions->useRaLocationsOption->isEnabled(),
```
DRvanR: same here as in the other PR, use JsonSerializable please.

src/Surfnet/StepupMiddleware/ApiBundle/Controller/InstitutionConfigurationOptionsController.php
```diff
@@ -0,0 +1,52 @@
...
+
+final class InstitutionConfigurationOptionsController extends Controller
+{
+    public function getForInstitutionAction($institution)
+    {
+        $this->denyAccessUnlessGranted(['ROLE_SS', 'ROLE_RA']);
+
+        $institutionConfigurationOptions = $this
+            ->getInstitutionConfigurationOptionsService()
+            ->findInstitutionConfigurationOptionsFor(new Institution($institution));
```
DRvanR: what happens if there is not InstitutionConfigurationOptions?

src/Surfnet/StepupMiddleware/ApiBundle/Resources/config/routing.yml
```diff
@@ -143,3 +143,11 @@ ra_location:
     requirements:
         raLocationId: ".+"
     condition: "request.headers.get('Accept') matches '/^application\\\\/json($|[;,])/'"
+
+institution_configuration_options_for_institution:
+    path:    /institution-configuration-options/{institution}
```
DRvanR: no sure how I feel about this URL... not sure that I believe that a resource should exist for an institution configuration options... especially with keeping ra-locations in mind. Dunno, it just feels off... can't exactly think of something obviously better, but yeah...

arothuis: I agree with this. Perhaps we offer an `GET \institution-configuration\{institution}`, which contains the options and the RA locations? 

It feels less weird to me: the options and the RA locations are related to the concept of configuration regarding an institution. However, they _are_ separate concepts. Also, the RA locations already have a separate API, so I guess you'd either have two ways to get RA locations or change the RA location API to fall under `\institution-configuration\`.

DRvanR: no, those (options and ralocations) are separate concepts, that should be treated separately.

was more thinking along the lines of "queryability" - using Institution as a get parameter, as in searching. 

`\institution-configuration\{institution}\ra-location` and `\institution-configuration\{institution}\options` kinda makes sense, but both still require institution to be a resource, with ra-locations and options being a sub resource....

Nevermind, keep it as is, not worth the hassle :wink:

src/Surfnet/StepupMiddleware/ApiBundle/Controller/InstitutionConfigurationOptionsController.php
arothuis: The InstitutionConfigurationOptionsService calls the InstitutionConfigurationOptionsRepository which does a `getSingleResult()` to return the results. 

This will throw a `NoResultException` when no result is found and a `NonUniqueResultException` if more than one are found. These will be converted to 500s by the `ExceptionListener`.

In the latter case, I'd say that is to be expected. It is not the client's fault the server has reached an invalid state with duplicate entries. Regarding the `NoResultException`, I think it should be converted to a `BadRequestException` informing the client no options could be found for institution "%s"/

arothuis:
@DRvanR

src/Surfnet/StepupMiddleware/ApiBundle/Controller/InstitutionConfigurationOptionsController.php
DRvanR: was a rhetorical question, as in the code that followed would have thrown a fatal "trying to access property of non-object" etc ;)

src/Surfnet/StepupMiddleware/ApiBundle/Controller/InstitutionConfigurationOptionsController.php
```diff
@@ -0,0 +1,53 @@
...
+
+        if ($institutionConfigurationOptions === null) {
+            throw new BadApiRequestException([
```
DRvanR: Nope, this should just be a 404. The resource does not exist, therefor it should be a 404 not found.

DRvanR:
with the 404 implemented :+1:

DRvanR:
oh and please create a tech debt issue that controllers should be converted services rather than using the container to retrieve whatever they  please :cat:

arothuis:
Controllers as services, yes please! Makes 'em a bit more testable than mere web test cases.

src/Surfnet/StepupMiddleware/ApiBundle/Controller/InstitutionConfigurationOptionsController.php
arothuis: Ah, of course 🔔

arothuis:
Note: after using the JsonSerializable behaviour on the InstitutionConfigurationOptions, the format has been aligned with how showing all options works in the Management API, for example:

```
GET /institution-configuration-options/dev.organisation.example

{
    "dev.organisation.example" : {
        "use_ra_locations": false,
        "show_raa_contact_information": true
    }
}
```

PR #151, opened 2016-08-04
Remove institution from show institution configuration options API

arothuis:
This way, the results from the "GET" api can be used and altered when POSTing.

When we decided to use the JSON serialized variant for the InstitutionConfigurationOptions, we missed this.


DRvanR:
:+1:

PR #152, opened 2016-08-05
Specify event listeners for replaying

arothuis:
This also contains a bugfix for the `InstitutionConfigurationProcessor` and a bit of clean-up.


src/Surfnet/StepupMiddleware/CommandHandlingBundle/DependencyInjection/CompilerPass/AddEventBusListenersCompilerPass.php
```diff
@@ -29,8 +30,31 @@ public function process(ContainerBuilder $container)
         $definition = $container->getDefinition('surfnet_stepup_middleware_command_handling.event_bus.buffered');
         $eventListenerDefinitions = $container->findTaggedServiceIds('event_bus.event_listener');
 
-        foreach (array_keys($eventListenerDefinitions) as $id) {
-            $definition->addMethodCall('subscribe', [new Reference($id)]);
+        // When replaying events, certain listeners might not be interested in acting upon, for instance
+        // when they are no longer relevant at the time of replaying (i.e. sending emails)
```
DRvanR: Nit: this is more about "should not be allowed to run again" than "might not be interested". It is really important that they don't run again - they might even cause crashes (attempting to create a second Inst.Conf.).

DRvanR:
one minor thing, other than that :+1:

DRvanR:
:+1:

PR #153, opened 2016-08-05
If showRaaContactInformation is false, RAAs should not be shown

arothuis:
This should be tested more extensively.


DRvanR:
:+1:

DRvanR:
:+1:

PR #157, opened 2016-10-24
Clear entity managers references after each event

arothuis:
[132894351](https://www.pivotaltracker.com/story/show/132894351)

Doctrine uses PHP's [spl_object_hash](http://php.net/manual/en/function.spl-object-hash.php). Hashes can be re-assigned after a certain object has been destroyed / garbage collected.

It seems that we have reached an edge case wherein hash collisions occur during event replays. Similar edge cases have been reported at the [Doctrine repository](https://github.com/doctrine/doctrine2/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aopen%20spl_object_hash%20).

Clearing the entity manager after each event ensures the entity manager correctly inserts entities that have not been seen before.

This does come with the caveat that event listeners cannot hold references to entities crossing events.


rjkip:
It is what it is. :+1:

PR #160, opened 2016-11-25
RAs should not be able to vet outside institution

arothuis:
This PR makes it impossible for RAs to vet registrants outside of their institution, except for SRAAs.

This PR also enforces that only Identities with RA privileges can vet second factors.

This relates to SURFnet/Stepup-RA#133.

PR #161, opened 2016-12-01
Revert RA-checking and cross-institution vetting check in Identity aggregate

arothuis:
This reverts commit a1543deca6a1f989b12eca2c119644b3fa951a6a, reversing
changes made to 4b0c8a62ef1182754d887529d2f98ca4f20ae291.

In the Identity aggregate, it is not known whether a certain identity is an SRAA or not. If he is an SRAA, the registrationAuthority property is set to null -- contrary to what one might assume.
That is why these checks currently do not work.

In the future, we need to find another way to determine in the Identity aggregate whether the one doing the vetting is a registration authority and whether he is an SRAA.
See: https://www.pivotaltracker.com/story/show/135335407

Reverting this means that, in theory, any Identity can vet a second factor regardless of whether they are within the same institution.
However, these checks are already done by the front-ends: RA and SS.

This reverts work done in PR #160.

### DIFF
--- a/src/Surfnet/StepupMiddleware/CommandHandlingBundle/Tests/Identity/CommandHandler/IdentityCommandHandlerTest.php
+++ b/src/Surfnet/StepupMiddleware/CommandHandlingBundle/Tests/Identity/CommandHandler/IdentityCommandHandlerTest.php
@@ -27,7 +27,6 @@ use Surfnet\Stepup\DateTime\DateTime;
 use Surfnet\Stepup\Identity\Entity\ConfigurableSettings;
 use Surfnet\Stepup\Identity\Event\EmailVerifiedEvent;
 use Surfnet\Stepup\Identity\Event\GssfPossessionProvenEvent;
-use Surfnet\Stepup\Identity\Event\IdentityAccreditedAsRaEvent;
 use Surfnet\Stepup\Identity\Event\IdentityCreatedEvent;
 use Surfnet\Stepup\Identity\Event\IdentityEmailChangedEvent;
 use Surfnet\Stepup\Identity\Event\IdentityRenamedEvent;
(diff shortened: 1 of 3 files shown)


arothuis:
@rjkip

src/Surfnet/StepupMiddleware/CommandHandlingBundle/Tests/Identity/CommandHandler/IdentityCommandHandlerTest.php
```diff
@@ -1472,7 +1092,7 @@ public function an_identity_can_express_its_locale_preference()
      * @test
      * @group command-handler
      * @runInSeparateProcess
-     * @expectedException \Surfnet\StepupMiddleware\CommandHandlingBundle\Exception\UnsupportedLocaleException
+     * @expectedException Surfnet\StepupMiddleware\CommandHandlingBundle\Exception\UnsupportedLocaleException
```
rjkip: Would be nice to reintroduce this.

DRvanR:
Perhaps it should be two seperate commands for SRAA and RA(A) who both can be validated, verified and treated differently? It are (significant) different actions, SRAA may vet from different institution, RA(A) may not.

arothuis:
I am going to verify this manually for all roles concerned.

arothuis:
@DRvanR 

Modelling them like separate concepts sounds like a plan.
Would the RA front-end then fire a separate command when the user is logged in as SRAA?

DRvanR:
> Would the RA front-end then fire a separate command when the user is logged in as SRAA?

yes, which can be verified in the MW commandhandling pipeline and domain (where necessary)

arothuis:
Thanks, I will put this in the aforementioned issue.

arothuis:
Merging this as it fixes the issue at hand.
For future improvements, see: https://www.pivotaltracker.com/story/show/135335407

PR #164, opened 2017-01-30
Introduction of RemoveInstitutionConfigurationByUnnormalizedIdCommand

DRvanR:
https://www.pivotaltracker.com/story/show/135610963

### DIFF
--- a/src/Surfnet/StepupMiddleware/CommandHandlingBundle/Configuration/Command/RemoveInstitutionConfigurationByUnnormalizedIdCommand.php
+++ b/src/Surfnet/StepupMiddleware/CommandHandlingBundle/Configuration/Command/RemoveInstitutionConfigurationByUnnormalizedIdCommand.php
@@ -1,0 +1,34 @@
+<?php
+
+/**
+ * Copyright 2017 SURFnet bv
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


src/Surfnet/StepupMiddleware/CommandHandlingBundle/Configuration/Command/RemoveInstitutionConfigurationByUnnormalizedIdCommand.php
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Surfnet\StepupMiddleware\CommandHandlingBundle\Configuration\Command;
+
+use Surfnet\StepupMiddleware\CommandHandlingBundle\Command\AbstractCommand;
+use Surfnet\StepupMiddleware\CommandHandlingBundle\Command\ManagementExecutable;
+
+final class RemoveInstitutionConfigurationByUnnormalizedIdCommand extends AbstractCommand implements
```
arothuis: Nit, but it could be something to consider. When a command says "ByUnnormalizedId", I would personally expect its parameters to match this to make its requirements as explicit as possible. In the most extreme case, that would mean changing `$institution` to `$unnormalizedInstitutionConfigurationId`, but something like `$unnormalizedInstitution` would be fine as well given that we use institutions as a natural identifier for InstitutionConfigurations.

DRvanR: (partially) Agree with the sentiment, however the InstitutionConfigurations are all fully based on the Institution on both the read and the write side. The name of the command shows the full intent of the command given, and contains the data needed to execute the command. 
In general though, a command name does not necessarily tell what the contents of the command are - it gives intent on what the application should do with the data it contains, it is a special DTO - a DTO with intent :wink:

arothuis: Thanks for the explanation. Consistency is key, let's keep it this way.

PR #169, opened 2017-02-07
Implement InstitutionConfiguration migration

arothuis:
[135610963](https://www.pivotaltracker.com/story/show/135610963)

The InstitutionConfigurations are migrated in order to update existing InstitutionConfigurationIds to be handled case-insensitively.

### DIFF
--- a/app/config/config_dev_event_replay.yml
+++ b/app/config/config_dev_event_replay.yml
@@ -3,6 +3,3 @@ imports:
 
 swiftmailer:
     disable_delivery: true
-
-surfnet_stepup_middleware_command_handling:
-    warn_on_missing_email_template: false
--- a/src/Surfnet/StepupMiddleware/MiddlewareBundle/Console/Command/MigrateInstitutionConfigurationsCommand.php
+++ b/src/Surfnet/StepupMiddleware/MiddlewareBundle/Console/Command/MigrateInstitutionConfigurationsCommand.php
@@ -68,7 +68,7 @@ final class MigrateInstitutionConfigurationsCommand extends Command
         $connectionHelper = $container->get('surfnet_stepup_middleware_middleware.dbal_connection_helper');
         $provider         = $container->get('surfnet_stepup_middleware_middleware.institution_configuration_provider');
         $pipeline         = $container->get('pipeline');
-        $eventBus         = $container->get('surfnet_stepup_middleware_command_handling.event_bus.buffered');
+        $entityManager    = $container->get('doctrine.orm.middleware_entity_manager');
 
         // The InstitutionConfiguration commands require ROLE_MANAGEMENT
         // Note that the new events will not have any actor metadata associated with them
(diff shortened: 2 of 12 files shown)


src/Surfnet/StepupMiddleware/MiddlewareBundle/Console/Command/MigrateInstitutionConfigurationsCommand.php
```diff
@@ -35,5 +40,74 @@ protected function configure()
...
+
+        $tokenStorage     = $container->get('security.token_storage');
+        $connectionHelper = $container->get('surfnet_stepup_middleware_middleware.dbal_connection_helper');
+        $provider         = $container->get('surfnet_stepup_middleware_middleware.institution_configuration_provider');
+        $pipeline         = $container->get('pipeline');
+        $eventBus         = $container->get('surfnet_stepup_middleware_command_handling.event_bus.buffered');
+
+        // The InstitutionConfiguration commands require ROLE_MANAGEMENT
+        // Note that the new events will not have any actor metadata associated with them
+        $tokenStorage->setToken(new AnonymousToken('console', 'console', ['ROLE_MANAGEMENT']));
```
arothuis: This is necessary because otherwise the AuthenticationStage in the StagedPipeline would fail:
```
  [Symfony\Component\Security\Core\Exception\AuthenticationCredentialsNotFoundException]                                             
  The token storage contains no authentication token. One possible reason may be that there is no firewall configured for this URL. 
```

This happens because the InstitutionConfiguration commands, when checked for authorization, require a token to be present and require this token to have `ROLE_MANAGEMENT` as they are intended to be triggered from the Management API. Setting an (anonymous) token with that role mitigates that issue.

DRvanR: Makes sense, would change console to something more CLI focused though (as discussed irl)

DRvanR:
As discussed: where now the `$eventBus->flush()` is placed, the entityManager should be flushed and emptied for performance issues as well as to prevent the Doctrine UnitOfWork from messing up the sequence of actions.

arothuis:
🔔

DRvanR:
Seems legit, :+1:

app/config/config_dev_event_replay.yml
```diff
@@ -3,6 +3,3 @@ imports:
 
 swiftmailer:
     disable_delivery: true
-
-surfnet_stepup_middleware_command_handling:
-    warn_on_missing_email_template: false
```
DRvanR: shouldn't this also mean that some configuration should be removed from a bundle?

src/Surfnet/StepupMiddleware/MiddlewareBundle/Console/Command/MigrateInstitutionConfigurationsCommand.php
```diff
@@ -35,5 +40,80 @@ protected function configure()
...
+        $entityManager    = $container->get('doctrine.orm.middleware_entity_manager');
+
+        // The InstitutionConfiguration commands require ROLE_MANAGEMENT, AddRaLocation requires ROLE_RA
+        // Note that the new events will not have any actor metadata associated with them
+        $tokenStorage->setToken(
+            new AnonymousToken('cli.institution_configuration_migration', 'cli', ['ROLE_MANAGEMENT', 'ROLE_RA'])
+        );
+
+        $connectionHelper->beginTransaction();
+
```
DRvanR: Perhaps add a message indicating that the command is running?

app/config/config_dev_event_replay.yml
arothuis: These settings have already been removed in the past (2a7e965219188b29d14cabfea0fcabaca3d333af), but apparently they were not removed from the event_replay specific configuration files.

arothuis:
I've added a message to indicate (which part of) the command is running.

PR #170, opened 2017-02-07
Document that institutions are dealt with case-insensitively in configuration

arothuis:
[135610963](https://www.pivotaltracker.com/story/show/135610963)

The phrasing probably needs some work. Suggestions welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -388,20 +388,12 @@ conform [RFC 1035](https://www.ietf.org/rfc/rfc1035.txt).
 
 Institutions, used as a natural identifier within several projections, 
 are therefore identified case-insensitively.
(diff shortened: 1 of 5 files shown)


DRvanR:
Should be expanded with:
- the rationale (schacHomeOrganization is case insensitive as according to spec-insert link-)
- why this matters for us (SHO==institution, lots of projections use institution as natural key)
- what the implications are (had to migrate configurations, dev must be aware, database uses utf8_collate_ci so did not notice earlier)
- whatever else may come to mind.

Oh, and add it to a new section "Caveats" or "warnings" or... something that makes it stand out enough that you can point people to it.

arothuis:
🔔

PR #171, opened 2017-02-07
Introduce Allowed Second Factor List

arothuis:
[119815553](https://www.pivotaltracker.com/story/show/119815553)

PR #172, opened 2017-02-09
Update reconfigure command in Management API

arothuis:

Branched from #171


arothuis:
As future developers might forget updating these values, I'd opt for merely referring to the file in the Stepup-bundle. Good suggestion!

PR #174, opened 2017-02-09
Introduce AllowedSecondFactorListUpdatedEvent

arothuis:
[119815553](https://www.pivotaltracker.com/story/show/119815553)

PR #175, opened 2017-02-10
Update institution configuration aggregate, command handling, API

arothuis:

Branched from #174

Update: rebased on develop.

### DIFF
--- a/src/Surfnet/Stepup/Configuration/InstitutionConfiguration.php
+++ b/src/Surfnet/Stepup/Configuration/InstitutionConfiguration.php
@@ -91,23 +91,19 @@ class InstitutionConfiguration extends EventSourcedAggregateRoot implements Inst
      */
     public static function create(InstitutionConfigurationId $institutionConfigurationId, Institution $institution)
     {
-        $useRaLocationsOption            = UseRaLocationsOption::getDefault();
-        $showRaaContactInformationOption = ShowRaaContactInformationOption::getDefault();
-        $allowedSecondFactorList         = AllowedSecondFactorList::blank();
-
         $institutionConfiguration = new self;
         $institutionConfiguration->apply(
             new NewInstitutionConfigurationCreatedEvent(
                 $institutionConfigurationId,
                 $institution,
-                $useRaLocationsOption,
-                $showRaaContactInformationOption
+                UseRaLocationsOption::getDefault(),
+                ShowRaaContactInformationOption::getDefault()
             )
         );
         $institutionConfiguration->apply(new AllowedSecondFactorListUpdatedEvent(
             $institutionConfigurationId,
             $institution,
-            $allowedSecondFactorList
+            AllowedSecondFactorList::blank()
         ));
 
         return $institutionConfiguration;
@@ -123,22 +119,18 @@ class InstitutionConfiguration extends EventSourcedAggregateRoot implements Inst
             throw new DomainException('Cannot rebuild InstitutionConfiguration as it has not been destroyed');
         }
 
-        $useRaLocationsOption            = UseRaLocationsOption::getDefault();
-        $showRaaContactInformationOption = ShowRaaContactInformationOption::getDefault();
-        $allowedSecondFactorList         = AllowedSecondFactorList::blank();
-
         $this->apply(
             new NewInstitutionConfigurationCreatedEvent(
                 $this->institutionConfigurationId,
                 $this->institution,
-                $useRaLocationsOption,
-                $showRaaContactInformationOption
+                UseRaLocationsOption::getDefault(),
+                ShowRaaContactInformationOption::getDefault()
             )
         );
         $this->apply(new AllowedSecondFactorListUpdatedEvent(
             $this->institutionConfigurationId,
             $this->institution,
-            $allowedSecondFactorList
+            AllowedSecondFactorList::blank()
         ));
 
         return $this;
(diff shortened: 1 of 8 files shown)


src/Surfnet/Stepup/Tests/Configuration/Value/AllowedSecondFactorListTest.php
```diff
@@ -144,6 +144,76 @@ public function a_second_factor_not_on_the_allowed_second_factor_list_is_not_all
...
+            $isSameAllowedSecondFactorList,
+            'An allowed second factor list should equal a second factor list with the same values but it does not'
+        );
+    }
+
+    /**
+     * @test
+     * @group domain
+     */
+    public function an_allowed_second_factor_list_does_not_equal_an_allowed_second_factor_list_with_different_values()
```
DRvanR: This, the previous and the next test are a bit... incomplete. They're not covering all cases that might be important. E.g.: if I change the first test (of amount of elements) in the `equals` method to 
```
if (count($other->allowedSecondFactors) < count($this->allowedSecondFactors)) { bork() }
```
all tests still pass. However, the objects won't agree on whether or not they're equal (`$a->equals($b) !== $b->equals($a)`).

They're structurally ok, but not thorough enough. As an example on how we did this in other projects, check https://github.com/OpenConext/SamlValueObject/blob/develop/tests/OpenConext/Value/Saml/EntitySetTest.php

src/Surfnet/StepupMiddleware/CommandHandlingBundle/Configuration/CommandHandler/InstitutionConfigurationCommandHandler.php
```diff
@@ -85,6 +87,14 @@ public function handleReconfigureInstitutionConfigurationOptionsCommand(
             new ShowRaaContactInformationOption($command->showRaaContactInformationOption)
         );
 
+        $allowedSecondFactors = array_map(function ($allowedSecondFactor) {
```
DRvanR: Nit: I'd pull up the decoding of a list to the top of the method (before loading the aggregate) to get a bit more structure in the code, now we're mixing a mapping of values with the invocation of methods. Why not for the single-value object (the options above)? Because they're more readable (can be inlined in the method call) than a 3 line mapping between two method calls :wink:

src/Surfnet/StepupMiddleware/CommandHandlingBundle/Tests/Configuration/CommandHandler/InstitutionConfigurationCommandHandlerTest.php
```diff
@@ -220,6 +251,98 @@ public function show_raa_contact_information_option_is_changed_if_its_given_valu
...
+        $useRaLocationsOption            = UseRaLocationsOption::getDefault();
+        $showRaaContactInformationOption = ShowRaaContactInformationOption::getDefault();
+        $originalAllowedSecondFactorList = AllowedSecondFactorList::blank();
+
+        $secondFactorsToAllow = ['sms', 'yubikey'];
+
+        $allowedSecondFactors = array_map(function ($secondFactorType) {
+            return new SecondFactorType($secondFactorType);
+        }, $secondFactorsToAllow);
+        $allowedSecondFactorList = AllowedSecondFactorList::ofTypes($allowedSecondFactors);
```
DRvanR: Suggestion: remove the mapping in favor of simple instantiation:
```
$originalAllowedSecondFactorList = AllowedSecondFactorList::blank();
$updatedAllowedSecondFactorList  = AllowedSecondFactorList::ofTypes([new SecondFactorType('sms'), new SecondFactorType('yubikey')]);
```

Visually identifies the difference, less code that requires thinking about what it does and thus detracts from what we're actually doing here (testing that it all works).

src/Surfnet/StepupMiddleware/CommandHandlingBundle/Tests/Configuration/CommandHandler/InstitutionConfigurationCommandHandlerTest.php
```diff
@@ -220,6 +251,98 @@ public function show_raa_contact_information_option_is_changed_if_its_given_valu
...
+    }
+
+    /**
+     * @test
+     * @group command-handler
+     */
+    public function allowed_second_factor_list_is_not_changed_if_its_values_are_the_same_as_the_current_list()
+    {
+        $secondFactorsToAllow = ['sms', 'yubikey'];
+        $allowedSecondFactors = array_map(function ($secondFactorType) {
```
DRvanR: same here as above.

src/Surfnet/Stepup/Configuration/InstitutionConfiguration.php
```diff
@@ -110,14 +123,23 @@ public function rebuild()
             throw new DomainException('Cannot rebuild InstitutionConfiguration as it has not been destroyed');
         }
 
+        $useRaLocationsOption            = UseRaLocationsOption::getDefault();
+        $showRaaContactInformationOption = ShowRaaContactInformationOption::getDefault();
+        $allowedSecondFactorList         = AllowedSecondFactorList::blank();
```
arothuis: I mirrored it from the `::create` method for consistency. I think I can inline them in both.

src/Surfnet/StepupMiddleware/CommandHandlingBundle/Tests/Configuration/CommandHandler/InstitutionConfigurationCommandHandlerTest.php
arothuis: I did it this way, because the `$secondFactorsToAllow` variable is still used in the `$command->allowedSecondFactors` and I like them to point to the same value internally as this reduces bugs in testing code.
I will change the `array_map` to a mapping by hand by using array access. This keeps the advantage of both.

arothuis:
Rebased on develop. Added requested changes.

PR #176, opened 2017-02-10
Allowed second factor projections

arothuis:
[119815553](https://www.pivotaltracker.com/story/show/119815553)

Branched from #175, #174

PR #177, opened 2017-02-10
Management GET API

arothuis:

Branched from: #176, #175, #174


app/DoctrineMigrations/Version20170210140051.php
```diff
@@ -0,0 +1,39 @@
...
+ */
+class Version20170210140051 extends AbstractMigration
+{
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf(
```
DRvanR: don't think we break up this line in any other migration 😄

src/Surfnet/Stepup/Configuration/Value/AllowedSecondFactorMap.php
```diff
@@ -0,0 +1,70 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Surfnet\Stepup\Configuration\Value;
+
+use Surfnet\StepupBundle\Value\SecondFactorType;
+
+final class AllowedSecondFactorMap
```
DRvanR: I'm not sold on this concept. More importantly though: no matter the usage or implementation,, it is most certainly not part of the domain, it must be moved out of the domain.

src/Surfnet/StepupMiddleware/ApiBundle/Configuration/Service/AllowedSecondFactorListService.php
```diff
@@ -0,0 +1,66 @@
...
+    /**
+     * @return AllowedSecondFactorMap
+     */
+    public function getAllowedSecondFactorMap()
+    {
+        $allInstitutions = array_map(function (ConfiguredInstitution $configuredInstitution) {
+            return $configuredInstitution->institution;
+        }, $this->configuredInstitutionRepository->findAll());
+
+        $allowedSecondFactorMap = AllowedSecondFactorMap::mappedTo($allInstitutions);
```
DRvanR: If the map is used, it must be implemented differently. Here a map is created that isn't actually a map, it is a list. It doesn't map anything. Furthermore, we now map lazily: the mapped value remains empty for institutions that allow all SecondFactorTypes. The empty value is converted to a `::ofTypes([])` rather than using `::blank()` and that is only done on demand when reading, not upon creation.

If you want to use it, the factory method should be something like:
```
AllowedSecondFactorMap::from($this->allowedSecondFactorRepository->findAll());
```

Alternative map (including probably a bunch of typo's as I typed it out here :wink: ):

```
class AllowedSecondFactorMap
{
    private $mappedAllowedSecondFactors;

    /**
     * @param AllowedSecondFactor[] $allowedSecondFactors
     */
    public static function from($allowedSecondFactors)
    {
        $allowedSecondFactorMap = new self();
        foreach ($allowedSecondFactors as $allowedSecondFactor) {
            $allowedSecondFactorMap->add($allowedSecondFactor);
        }

        return $allowedSecondFactorMap;
    }

    public function getAllowedSecondFactorListFor(Institution $institution)
    {
        $institution = strtolower($allowedSecondFactor->institution->getInstution());
        if (!array_key_exists($institution, $this->mappedAllowedSecondFactors)) {
            return AllowedSecondFactorList::blank();
        }

        return AllowedSecondFactorList::ofTypes($this->mappedAllowedSecondFactors[$institution]);
    }

    private function __construct()
    {
    }

    private function add(AllowedSecondFactor $allowedSecondFactor)
    {
        $institution = strtolower($allowedSecondFactor->institution->getInstution());

        $this->mappedAllowedSecondFactors[$institution][] = $allowedSecondFactor->secondFactor;
    }
}
```

Simpler, always correct (!), and uses the methods of other VO's as intended (blank 😉 )

Thoughts?

src/Surfnet/StepupMiddleware/ManagementBundle/Controller/InstitutionConfigurationController.php
```diff
@@ -145,6 +151,14 @@ private function getInstitutionConfigurationOptionsService()
     }
 
     /**
+     * @return AllowedSecondFactorListService
+     */
+    private function getAllowedSecondFactorListService()
+    {
+        return $this->get('surfnet_stepup_middleware_api.service.allowed_second_factor_list');
```
DRvanR: ugh, gimme controllers as services already 😭

src/Surfnet/Stepup/Configuration/Value/AllowedSecondFactorMap.php
arothuis: This object was introduced to simplify the logic in the AllowedSecondFactorListService and to make it possible to return an AllowedSecondFactorList per institution. I'm not sure about alternative approaches, but I agree this solution should not be part of the domain.

src/Surfnet/StepupMiddleware/ApiBundle/Configuration/Service/AllowedSecondFactorListService.php
arothuis: I like this approach as it more tightly fits to the AllowedSecondFactorList and clearly conveys a blank list to be blank :1st_place_medal:

Not sure why I didn't build it that way, must have been a friday afternoon ;)

src/Surfnet/StepupMiddleware/ManagementBundle/Controller/InstitutionConfigurationController.php
arothuis: In [due time](https://www.pivotaltracker.com/story/show/139669445).

arothuis:
Pushed changes, thanks for the suggestions.
Will rebase to develop after #176 is merged.

arothuis:
Rebased on develop.

PR #179, opened 2017-02-13
Enforce Allowed Second Factor List

arothuis:

### DIFF
--- a/src/Surfnet/Stepup/Configuration/InstitutionConfiguration.php
+++ b/src/Surfnet/Stepup/Configuration/InstitutionConfiguration.php
@@ -172,6 +172,7 @@ class InstitutionConfiguration extends EventSourcedAggregateRoot implements Inst
 
     public function updateAllowedSecondFactorList(AllowedSecondFactorList $allowedSecondFactorList)
     {
+        // AllowedSecondFactorList can be null for InstitutionConfigurations for which this functionality did not exist
         if ($this->allowedSecondFactorList !== null
             && $this->allowedSecondFactorList->equals($allowedSecondFactorList)
         ) {
--- a/src/Surfnet/Stepup/Configuration/InstitutionConfiguration.php
+++ b/src/Surfnet/Stepup/Configuration/InstitutionConfiguration.php
@@ -172,7 +172,9 @@ class InstitutionConfiguration extends EventSourcedAggregateRoot implements Inst
 
     public function updateAllowedSecondFactorList(AllowedSecondFactorList $allowedSecondFactorList)
     {
-        if ($this->allowedSecondFactorList->equals($allowedSecondFactorList)) {
+        if ($this->allowedSecondFactorList !== null
+            && $this->allowedSecondFactorList->equals($allowedSecondFactorList)
+        ) {
             return;
         }
 
--- a/src/Surfnet/StepupMiddleware/ApiBundle/Configuration/Service/AllowedSecondFactorListService.php
+++ b/src/Surfnet/StepupMiddleware/ApiBundle/Configuration/Service/AllowedSecondFactorListService.php
@@ -20,7 +20,6 @@ namespace Surfnet\StepupMiddleware\ApiBundle\Configuration\Service;
 
 use Surfnet\Stepup\Configuration\Value\AllowedSecondFactorList;
 use Surfnet\Stepup\Configuration\Value\Institution;
-use Surfnet\StepupBundle\Value\SecondFactorType;
 use Surfnet\StepupMiddleware\ApiBundle\Configuration\Entity\AllowedSecondFactor;
 use Surfnet\StepupMiddleware\ApiBundle\Configuration\Repository\AllowedSecondFactorRepository;
 use Surfnet\StepupMiddleware\ApiBundle\Configuration\Repository\ConfiguredInstitutionRepository;
@@ -60,17 +59,5 @@ class AllowedSecondFactorListService
     public function getAllowedSecondFactorMap()
     {
         return AllowedSecondFactorMap::from($this->allowedSecondFactorRepository->findAll());
-    }
-
-    /**
-     * @param SecondFactorType $secondFactor
-     * @param Institution $institution
-     * @return bool
-     */
-    public function isSecondFactorAllowedFor(SecondFactorType $secondFactor, Institution $institution)
-    {
-        $allowedSecondFactorList = $this->getAllowedSecondFactorListFor($institution);
-
-        return $allowedSecondFactorList->allows($secondFactor);
     }
 }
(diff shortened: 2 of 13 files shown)


src/Surfnet/Stepup/Configuration/InstitutionConfiguration.php
```diff
@@ -172,7 +172,9 @@ public function configureShowRaaContactInformationOption(ShowRaaContactInformati
 
     public function updateAllowedSecondFactorList(AllowedSecondFactorList $allowedSecondFactorList)
     {
-        if ($this->allowedSecondFactorList->equals($allowedSecondFactorList)) {
+        if ($this->allowedSecondFactorList !== null
```
DRvanR: It should be documented that this can only be `null` due to pre-existing InstitutionConfigurations where this functionality did not yet exist.

src/Surfnet/StepupMiddleware/ApiBundle/Configuration/Service/AllowedSecondFactorListService.php
```diff
@@ -60,4 +61,16 @@ public function getAllowedSecondFactorMap()
     {
         return AllowedSecondFactorMap::from($this->allowedSecondFactorRepository->findAll());
     }
+
+    /**
+     * @param SecondFactorType $secondFactor
+     * @param Institution $institution
+     * @return bool
+     */
+    public function isSecondFactorAllowedFor(SecondFactorType $secondFactor, Institution $institution)
```
DRvanR: what is the purpose of this method? Why is a decision, that is deliberately exposed in a VO, hidden in a service? I would have expected the `AllowedSecondFactorList` to be retrieved with the "getter" and then used for the decision. This seems to hide the complexity for no apparent gain? Thoughts?

src/Surfnet/StepupMiddleware/CommandHandlingBundle/EventListener/CommandProcessingExceptionListener.php
```diff
@@ -18,13 +18,13 @@
...
 
-use Surfnet\StepupMiddleware\CommandHandlingBundle\Pipeline\Exception\InvalidCommandException;
+use Surfnet\StepupMiddleware\CommandHandlingBundle\Pipeline\Exception\ProcessingAbortedException;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpKernel\Event\GetResponseForExceptionEvent;
 use Symfony\Component\HttpKernel\KernelEvents;
 
-class InvalidCommandExceptionListener implements EventSubscriberInterface
+class CommandProcessingExceptionListener implements EventSubscriberInterface
```
DRvanR: Will this handle all CommandProcessingExceptions? I.e. when a DomainException is thrown? Which exceptions can I expect it to handle?

DRvanR: As an aside: what would you say about making a tech debt issue that allows for catching this specific exception at the controller level? Seems like something the controller would do. Would make it more apparent that this exists and handled...

src/Surfnet/StepupMiddleware/CommandHandlingBundle/Exception/SecondFactorNotAllowedException.php
```diff
@@ -0,0 +1,45 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Surfnet\StepupMiddleware\CommandHandlingBundle\Exception;
+
+use Exception;
```
DRvanR: Seems this is obsolete?

src/Surfnet/StepupMiddleware/CommandHandlingBundle/Identity/CommandHandler/IdentityCommandHandler.php
```diff
@@ -278,4 +300,20 @@ private function assertIsValidLocale(Locale $locale)
...
     }
+
+    private function assertSecondFactorIsAllowedFor(SecondFactorType $secondFactor, Institution $institution)
+    {
+        $isSecondFactorAllowed = $this->allowedSecondFactorListService->isSecondFactorAllowedFor(
+            $secondFactor,
+            new ConfigurationInstitution($institution->getInstitution())
+        );
+
+        if ($isSecondFactorAllowed === false) {
```
DRvanR: With not "hiding" the call, this would become `if (!$allowedSecondFactorList->allows($secondFactorType)) {` which to me seems a better indicator...

src/Surfnet/StepupMiddleware/CommandHandlingBundle/Tests/Identity/CommandHandler/IdentityCommandHandlerTest.php
```diff
@@ -74,17 +77,25 @@ class IdentityCommandHandlerTest extends CommandHandlerTest
...
+    /**
+     * @var AllowedSecondFactorListService|m\MockInterface
+     */
+    private $mockAllowedSecondFactorListService;
+
     protected function createCommandHandler(EventStoreInterface $eventStore, EventBusInterface $eventBus)
     {
         $aggregateFactory = new PublicConstructorAggregateFactory();
 
+        $this->mockAllowedSecondFactorListService = m::mock(AllowedSecondFactorListService::class);
```
DRvanR: there's no `setup` method to use? Seems like a strange place to do this...

src/Surfnet/StepupMiddleware/ApiBundle/Configuration/Service/AllowedSecondFactorListService.php
arothuis: You are correct. It's a force of habit that I want to hide everything behind a service call, but we have an intelligent (domain) object for this. Let's use that one instead.

src/Surfnet/StepupMiddleware/CommandHandlingBundle/EventListener/CommandProcessingExceptionListener.php
arothuis: I thought it should handle all Exceptions related to command processing. On the other hand, this listener is specifically for the  `ProcessingAbortedException`, so maybe match that.

Regarding the tech debt issue. Although I'm not sure if I follow what you are saying, I think we're on the same page. I would expect any exception as a result of an API call to be caught at controller level where its output is to be converted to API-specific response data not in some (hidden) listener.

arothuis:
Some updates ready for review.

PR #180, opened 2017-02-15
Introduce specific event replay command

arothuis:

This new command is based on the ReplayEventsCommand that runs a replayer for all events, which was primarily used for development purposes.
Its purpose is to replay specific events for specific projectors.

### DIFF
--- a/src/Surfnet/StepupMiddleware/MiddlewareBundle/Console/Command/ReplaySpecificEventsCommand.php
+++ b/src/Surfnet/StepupMiddleware/MiddlewareBundle/Console/Command/ReplaySpecificEventsCommand.php
@@ -30,7 +30,7 @@ class ReplaySpecificEventsCommand extends Command
     protected function configure()
     {
         $this
-            ->setName('middleware:event:specific_replay')
+            ->setName('stepup:event:replay')
             ->setDescription('replay specified events for specified projectors');
     }
 
--- a/src/Surfnet/StepupMiddleware/MiddlewareBundle/Console/Command/ReplaySpecificEventsCommand.php
+++ b/src/Surfnet/StepupMiddleware/MiddlewareBundle/Console/Command/ReplaySpecificEventsCommand.php
@@ -19,11 +19,8 @@
 namespace Surfnet\StepupMiddleware\MiddlewareBundle\Console\Command;
 
 use Symfony\Component\Console\Command\Command;
-use Symfony\Component\Console\Helper\FormatterHelper;
-use Symfony\Component\Console\Helper\QuestionHelper;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
-use Symfony\Component\Console\Question\ConfirmationQuestion;
 
 class ReplaySpecificEventsCommand extends Command
 {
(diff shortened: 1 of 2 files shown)


src/Surfnet/StepupMiddleware/MiddlewareBundle/Console/Command/ReplaySpecificEventsCommand.php
```diff
@@ -0,0 +1,72 @@
...
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\ConfirmationQuestion;
+
+class ReplaySpecificEventsCommand extends Command
+{
+    protected function configure()
+    {
+        $this
+            ->setName('middleware:event:specific_replay')
```
arothuis: Not sure about the name. I would prefer `middleware:event:replay`, but that one is already taken for the deprecated/development replay command. I could rename or remove that one, but that feels odd when there is no replacement in place.

Thoughts?

DRvanR: `stepup:events:replay` (domain:component-of-action:action) - kinda aligns with the `stepup:migrate:institution-configuration`

I feel using the `stepup:` as new base NS is a good differentiator of intent with the existing `middleware` ones.

src/Surfnet/StepupMiddleware/MiddlewareBundle/Console/Command/ReplaySpecificEventsCommand.php
```diff
@@ -0,0 +1,72 @@
...
+        $formatter = $this->getHelper('formatter');
+
+        /** @var QuestionHelper $interrogator */
+        $interrogator = $this->getHelper('question');
+
+        $output->writeln(['', $formatter->formatBlock([
+            '',
+            'WARNING!!!!',
+            '',
+            'You are about to WIPE all read data and recreate all data based on chosen events and projectors.',
```
DRvanR: This should plain simply not be true. As discussed yesterday - this is intended to be a proper event replay, through specific replay-projectors.

In general: please remove the copy pasted code. For now just a message that it does not actually do anything should suffice :wink:

src/Surfnet/StepupMiddleware/MiddlewareBundle/Console/Command/ReplaySpecificEventsCommand.php
arothuis: Good idea.

src/Surfnet/StepupMiddleware/MiddlewareBundle/Console/Command/ReplaySpecificEventsCommand.php
arothuis: OK.

arothuis:
I have removed all `execute()` code for now and moved it to `stepup:event:replay`. 

Do note that the command name is still `ReplaySpecificEvents` as it would conflict with `ReplayEvents`. IMHO, this is fine.

PR #181, opened 2017-02-15
Introduce event collection and events.yml

arothuis:

Branched from: #180

### DIFF
--- a/src/Surfnet/StepupMiddleware/MiddlewareBundle/EventSourcing/EventCollection.php
+++ b/src/Surfnet/StepupMiddleware/MiddlewareBundle/EventSourcing/EventCollection.php
@@ -25,7 +25,7 @@ final class EventCollection
     /**
      * @var string[]
      */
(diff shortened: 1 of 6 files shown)


src/Surfnet/StepupMiddleware/MiddlewareBundle/EventSourcing/EventCollection.php
```diff
@@ -0,0 +1,40 @@
...
+{
+    /**
+     * @var string[]
+     */
+    public $eventNames = [];
+
+    public function __construct($eventNames)
+    {
+        foreach ($eventNames as $eventName) {
+            if (!is_string($eventName) || empty($eventName)) {
```
DRvanR: could check with `class_exists` whether or not the event actually exists.

arothuis: Nice plan!

PR #188, opened 2017-02-16
Add projector selection to ReplaySpecificEventsCommand 

arothuis:

### DIFF
--- a/src/Surfnet/StepupMiddleware/MiddlewareBundle/EventSourcing/ProjectorCollection.php
+++ b/src/Surfnet/StepupMiddleware/MiddlewareBundle/EventSourcing/ProjectorCollection.php
@@ -69,7 +69,7 @@ final class ProjectorCollection implements IteratorAggregate
                 );
             }
 
-            $subsetCollection->projectors[$projectorName] = $this->projectors;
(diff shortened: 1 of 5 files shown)


src/Surfnet/StepupMiddleware/MiddlewareBundle/EventSourcing/ProjectorCollection.php
```diff
@@ -29,9 +30,58 @@
...
+    /**
+     * @return string[]
+     */
+    public function getProjectorNames()
+    {
+        return array_map(
+            function (ProjectorInterface $projector) {
+                return get_class($projector);
+            },
+            array_values($this->projectors)
```
DRvanR: why `array_values`? Isn't that the case anyways with a map?

arothuis: Let me check, but I thought `array_map` kept the keys as well. That would mean you would have to type out the projector names in full.

arothuis: It keeps the keys in all PHP versions:
https://3v4l.org/VI6ZQ

PR #189, opened 2017-02-16
Document adding events

arothuis:


arothuis:
I think that is a good idea. I'll create a chore for it.

To go even further, I wonder if it would be possible to find events not registered in `events.yml`.
That way, you would be able to test if all requirements for creating an event are met. I think it should be possible, but it might be too expensive for our test suite.

PR #190, opened 2017-02-17
Introduce specific dispatcher

arothuis:
[133173121](https://www.pivotaltracker.com/story/show/133173121)

This PR merely introduces the concept of a specific event dispatcher that can be called from the Replay CLI command, using a given EventCollection and ProjectorCollection. I've outlined the planned approach in the comments in the try/catch block.

Branched from #188

PR #191, opened 2017-02-17
Dispatch events in ReplaySpecificEventsCommand

arothuis:

Branched from #190.


arothuis:
We could use some more IO, I think, for example:
- [ ] a confirmation question after the user has selected events and projectors
- [ ] some messages to tell the user what is going on.

arothuis:
Separated concerns, provided minimal IO. If more output is required, a manager can be created which is responsible for the order of operations and giving feedback in between chunks of events.

PR #192, opened 2017-02-21
Fix tests for PHP 7.1

nicwortel:


arothuis:
Not sure why the tests fail in the CI and it will be difficult to debug (unless you can reproduce it or make the tests spew the error response when they're failing).

Other than that, forward compatibility is nice! Do note that it is not a goal in itself as we currently support 5.6 and 7.0.

For future reference: this issue arises from [DateTime incorporating microseconds since 7.1](http://php.net/manual/en/migration71.incompatible.php#migration71.incompatible.datetime-microseconds).

PR #194, opened 2017-02-21
Filter second factor types

nicwortel:
Allow to filter on second factor types, in order to request a list of RA candidates with a specific LoA.

https://www.pivotaltracker.com/story/show/131617617


arothuis:
Please add a reference to the Pivotal issue.

PR #195, opened 2017-03-07
Remove Guzzle as a direct dependency, update Stepup-bundle

arothuis:
[126126187](https://www.pivotaltracker.com/story/show/126126187)

See also: b36bc01c1f6da503aa88bf3325dcea26250d3809

**All Guzzle updates:**
https://github.com/OpenConext/Stepup-bundle/pull/40
https://github.com/SURFnet/yubikey-api-client/pull/5
https://github.com/SURFnet/messagebird-api-client-bundle/pull/14
https://github.com/OpenConext/Stepup-Middleware-clientbundle/pull/51

https://github.com/OpenConext/Stepup-Gateway/pull/113
https://github.com/OpenConext/Stepup-RA/pull/136
https://github.com/OpenConext/Stepup-SelfService/pull/121
https://github.com/OpenConext/Stepup-Middleware/pull/195

An issue for installing the QA-tools (v3) and reconfiguring the pre-commit-hook [has been added](https://www.pivotaltracker.com/story/show/141134485).


nicwortel:
I see that the "guzzlehttp/guzzle" is removed from "composer.json", but it is not removed from "composer.lock". This probably means that some other package is still requiring Guzzle. Is this intended?

PR #198, opened 2017-08-21
Add a 'use PDP' flag to IdP/Sp configuration

jorissteyn:
The flag 'use_pdp' can be omitted and defaults to false.

### DIFF
--- a/docs/MiddlewareConfiguration.md
+++ b/docs/MiddlewareConfiguration.md
@@ -180,7 +180,7 @@ Each element in the ```service_providers``` array must be an object and contain 
 * `second_factor_only_nameid_patterns` contains a list of patterns (strings that may contain a '*' wildcard character) that are allowed to use the Second Factor Only mode. E.g. the wilcard pattern `urn:collab:person:example.org:*` matches all NameIDs that start with "urn:collab:person:example.org:". Does nothing if `second_factor_only` is not set to true.
 * ```assertion_encryption_enabled``` must be a boolean value that allows configuring whether or not the assertion that is sent to the SP should be encrypted.
 * ```blacklisted_encryption_algorithms``` contains an array that lists (each as single string-element) algorithms that may not be used for encryption. When left empty, no algorithms are blacklisted. As the gateway currently only allows the " http://www.w3.org/2001/04/xmldsig-more#rsa-sha256" algorithm this option is of little practical use.
-
+* `use_pdp` optional boolean value, defaults to false, the PDP policy decision is enforced when enabled
 
 #### Identity Providers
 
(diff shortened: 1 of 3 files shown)


docs/MiddlewareConfiguration.md
```diff
@@ -180,6 +180,7 @@ Each element in the ```service_providers``` array must be an object and contain
 * `second_factor_only_nameid_patterns` contains a list of patterns (strings that may contain a '*' wildcard character) that are allowed to use the Second Factor Only mode. E.g. the wilcard pattern `urn:collab:person:example.org:*` matches all NameIDs that start with "urn:collab:person:example.org:". Does nothing if `second_factor_only` is not set to true.
 * ```assertion_encryption_enabled``` must be a boolean value that allows configuring whether or not the assertion that is sent to the SP should be encrypted.
 * ```blacklisted_encryption_algorithms``` contains an array that lists (each as single string-element) algorithms that may not be used for encryption. When left empty, no algorithms are blacklisted. As the gateway currently only allows the " http://www.w3.org/2001/04/xmldsig-more#rsa-sha256" algorithm this option is of little practical use.
+* `use_pdp` optional boolean value, defaults to false, if enabled the PDP endpoint will be called and the required LoA increased if the policy decision obligates a higher LoA
```
MKodde: This sentence does not flow very well. The same goes for the Identity Providers section below.

src/Surfnet/StepupMiddleware/ManagementBundle/Validator/IdentityProviderConfigurationValidator.php
```diff
@@ -27,22 +27,29 @@ public function validate(array $configuration, $propertyPath)
     {
         Assertion::isArray($configuration, 'invalid configuration format, must be an object', $propertyPath);
 
-        $acceptedProperties = [
+        $requiredProperties = [
```
MKodde: In my opinion, the newly chosen name is a step back in descriptiveness. The `use_pdp` property is not really required as it defaults to false when it is not configured. This also goes for the `ServiceProviderConfigurationValidator`.

jorissteyn: As discussed, use_pdp is optional (and removed from the list) so all params in $acceptedProperties are now required -> hence requiredProperties.

docs/MiddlewareConfiguration.md
jorissteyn: Thanks, rewrote it to "the PDP policy decision is enforced when enabled"

src/Surfnet/StepupMiddleware/ManagementBundle/Validator/IdentityProviderConfigurationValidator.php
```diff
@@ -27,22 +27,28 @@ public function validate(array $configuration, $propertyPath)
...
             'entity_id',
             'loa',
```
MKodde: We need the `use_pdp` configuration parameter in the list of `requiredProperties`. This because the `StepupAssert::keysMatch` throws an exception when the `use_pdp` configuration is handled.

I suggest to add `use_pdp` to the list of `requiredProperties`. 

This also goes for the `IdPConfigurationValidator`

MKodde: I have changed this for now, as we want to release this to the testing environment. Feel free to add to the discussion @jorissteyn or change this to a solution that is more to your liking.

jorissteyn: I see how we broke this now, merging this one.

PR #211, opened 2018-01-24
Add validation on institution SHO

MKodde:
The institution identifier (matching the SHO for authenticating users)
must be set as lower case values. This is validated in the assertion
method that was added to the ServiceProviderConfigurationValidator.


src/Surfnet/StepupMiddleware/ManagementBundle/Validator/ServiceProviderConfigurationValidator.php
```diff
@@ -137,6 +137,13 @@ private function validateLoaDefinition($configuration, $propertyPath)
         Assertion::isArray($value, 'must be an object', $path);
         Assertion::keyExists($value, '__default__', "must have the default loa set on the '__default__' property", $path);
         Assertion::allString($value, 'all properties must contain strings as values', $path);
+
+        // Test if all SP specific LoA configuration entries are lower case.
+        $this->assertValidInstitutionIdentifiers(
+            $value,
+            'all non __default__ properties must have a valid SHO',
```
jorissteyn: It might be more helpful to say something like "The ShacHomeOrganisation names in SP LoA configuration must all be lower case".

MKodde: Thanks, updated.

PR #221, opened 2018-04-18
Skip transition of unverified -> verified when email is disabled (new "proven" event)

jorissteyn:
Institutions which do not require email verification found 'Email
verified' events in the audit log.

Previous implementation was internally creating an unverified second
factor and using the verify-email event to transition it to verified,
causing this unwanted effect. The new implementation skips the
creationg of unverified second factors altogether.

So, with email verification:

    user proves possession -> create unverified -> user verifies email -> transition to verified

Previously, without email verification

    user proves possession -> create unverified -> transition to verified

New implementation, without email verification

    user proves possession -> create verified

This is achieved by creating new 'proven and verified' events for
tokens which are implicitly verified by the Idp. These events are
handled differently from the regular 'proven' events, and to maintain
compatibility with the previous way of working, handling of those
regular events is unchanged.

PR #222, opened 2018-04-18
Skip transition of unverified -> verified when email is disabled (experimental w/versioning)

jorissteyn:
Institutions which do not require email verification found 'Email
verified' events in the audit log.

Previous implementation was internally creating an unverified second
factor and using the verify-email event to transition it to verified,
causing this unwanted effect. The new implementation skips the
creationg of unverified second factors altogether.

So, with email verification:

    user proves possession -> create unverified -> user verifies email -> transition to verified

Previously, without email verification

    user proves possession -> create unverified -> transition to verified

New implementation, without email verification

    user proves possession -> create verified

This is achieved by creating a new 'ImplicitlyVerifiedByIdp' event,
which acts just like the 'EmailVerified' event but does not implement
the Auditable interface, thus not ending up in the audit log.

See Identity::applyGssfPossessionProvenEvent for an explanation of
what's wrong with this solution. Also see: applyEmailVerifiedEvent().


MKodde:
Closing this PR in favor of the solution proposed in #221 which is better in more than one way.

PR #223, opened 2018-05-02
Always send registration confirmation email

jorissteyn:
PR #221 broke the registration confirmation email for institutions
with email verification disabled. Stepup now always sends the
confirmation email regardless of the email verification setting.

### DIFF
--- a/src/Surfnet/Stepup/Identity/Event/GssfPossessionProvenAndVerifiedEvent.php
+++ b/src/Surfnet/Stepup/Identity/Event/GssfPossessionProvenAndVerifiedEvent.php
@@ -25,13 +25,14 @@ use Surfnet\Stepup\Identity\Value\Email;
 use Surfnet\Stepup\Identity\Value\GssfId;
 use Surfnet\Stepup\Identity\Value\IdentityId;
 use Surfnet\Stepup\Identity\Value\Institution;
+use Surfnet\Stepup\Identity\Value\Locale;
 use Surfnet\Stepup\Identity\Value\SecondFactorId;
 use Surfnet\Stepup\Identity\Value\StepupProvider;
 use Surfnet\StepupBundle\Value\SecondFactorType;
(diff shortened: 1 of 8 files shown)


src/Surfnet/Stepup/Identity/Event/GssfPossessionProvenAndVerifiedEvent.php
```diff
@@ -115,6 +124,11 @@ public function getAuditLogMetadata()
 
     public static function deserialize(array $data)
     {
+        // BC compatibility for test-environment only (2.8.0, fixed in 2.8.1)
+        if (!isset($data['preferred_locale'])) {
```
MKodde: It's really nice of you to include this piece of code. However, it's very installation dependent, if for example another Stepup user with entirely different locale settings would not benefit from this. If you could inject the `default_locale` into this event..

src/Surfnet/Stepup/Identity/Event/PhonePossessionProvenAndVerifiedEvent.php
```diff
@@ -106,13 +115,19 @@ public function getAuditLogMetadata()
 
     public static function deserialize(array $data)
     {
+        // BC compatibility for test-environment only (2.8.0, fixed in 2.8.1)
+        if (!isset($data['preferred_locale'])) {
+            $data['preferred_locale'] = 'en_GB';
```
MKodde: See previous comment.

src/Surfnet/Stepup/Identity/Event/GssfPossessionProvenAndVerifiedEvent.php
MKodde: We discussed this IRL, this feature is for the benefit of event replays. In event replay context, no email is sent. So the value of 'preferred_locale' is irrelevant. For the sake of efficiency we've set the value with the Surfnet default.

PR #232, opened 2018-09-18
Update the institution configuration aggregate

MKodde:
**What's new?**
 * Added support for configuring the three new FGA options (useRa, useRaa, selectRaa)
 * Updated the validator, also testing the validity of the institutions against the whitelist
 * Specific tests have been added to test these points

**What's not yet working**
No projections have been touched yet. So this breaks with the 'direct consistent' rule stated in the middleware wiki. The new options are not projected directly yet. This will be taken care of in the next PR. Be sure you remove the events that are fired during test of this PR.

The showAction (InstitutionConfigurationController) was not touched yet. And it makes more sense to address this when updating the projections.

**How to test?**
Personally, I've used Postman to test the configuration post on the Middleware API. Combined with XDebug & monitoring of the log file

Updated Postman payload:
```
{
  "institution-a.example.com": {
    "use_ra_locations": true,
    "show_raa_contact_information": true,
    "verify_email": true,
    "allowed_second_factors": [],
    "number_of_tokens_per_identity": 2,
    "use_ra": ["institution-a.example.com"],
    "use_raa": ["institution-b.example.com"]
    "select_raa": ["institution-b.example.com"]
  }
}
```

Test variations of the new options. For example leave them out (they are optional after all) or provide an empty array as a value.

See: https://www.pivotaltracker.com/story/show/160546629

### DIFF
--- a/src/Surfnet/Stepup/Tests/Configuration/Event/EventSerializationAndDeserializationTest.php
+++ b/src/Surfnet/Stepup/Tests/Configuration/Event/EventSerializationAndDeserializationTest.php
@@ -47,8 +47,11 @@ use Surfnet\Stepup\Configuration\Value\Location;
 use Surfnet\Stepup\Configuration\Value\NumberOfTokensPerIdentityOption;
 use Surfnet\Stepup\Configuration\Value\RaLocationId;
 use Surfnet\Stepup\Configuration\Value\RaLocationName;
+use Surfnet\Stepup\Configuration\Value\SelectRaaOption;
 use Surfnet\Stepup\Configuration\Value\ShowRaaContactInformationOption;
+use Surfnet\Stepup\Configuration\Value\UseRaaOption;
 use Surfnet\Stepup\Configuration\Value\UseRaLocationsOption;
+use Surfnet\Stepup\Configuration\Value\UseRaOption;
 use Surfnet\Stepup\Configuration\Value\VerifyEmailOption;
 use Surfnet\StepupBundle\Value\SecondFactorType;
 
@@ -124,7 +127,10 @@ class EventSerializationAndDeserializationTest extends TestCase
                     new UseRaLocationsOption(true),
                     new ShowRaaContactInformationOption(true),
                     new VerifyEmailOption(true),
-                    new NumberOfTokensPerIdentityOption(0)
+                    new NumberOfTokensPerIdentityOption(0),
+                    new UseRaOption(null),
+                    new UseRaaOption(null),
+                    new SelectRaaOption(null)
                 )
             ],
             'UseRaLocationsOptionChangedEvent' => [
--- a/src/Surfnet/Stepup/Tests/Configuration/InstitutionConfigurationTest.php
+++ b/src/Surfnet/Stepup/Tests/Configuration/InstitutionConfigurationTest.php
@@ -28,8 +28,11 @@ use Surfnet\Stepup\Configuration\Value\AllowedSecondFactorList;
 use Surfnet\Stepup\Configuration\Value\Institution;
 use Surfnet\Stepup\Configuration\Value\InstitutionConfigurationId;
 use Surfnet\Stepup\Configuration\Value\NumberOfTokensPerIdentityOption;
+use Surfnet\Stepup\Configuration\Value\SelectRaaOption;
 use Surfnet\Stepup\Configuration\Value\ShowRaaContactInformationOption;
+use Surfnet\Stepup\Configuration\Value\UseRaaOption;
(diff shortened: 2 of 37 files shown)


src/Surfnet/Stepup/Tests/Configuration/Event/EventSerializationAndDeserializationTest.php
```diff
@@ -47,8 +47,11 @@
 use Surfnet\Stepup\Configuration\Value\NumberOfTokensPerIdentityOption;
 use Surfnet\Stepup\Configuration\Value\RaLocationId;
 use Surfnet\Stepup\Configuration\Value\RaLocationName;
+use Surfnet\Stepup\Configuration\Value\SelectRaaOption;
 use Surfnet\Stepup\Configuration\Value\ShowRaaContactInformationOption;
+use Surfnet\Stepup\Configuration\Value\UseRaaOption;
 use Surfnet\Stepup\Configuration\Value\UseRaLocationsOption;
+use Surfnet\Stepup\Configuration\Value\UseRaOption;
 use Surfnet\Stepup\Configuration\Value\VerifyEmailOption;
 use Surfnet\StepupBundle\Value\SecondFactorType;
```
pablothedude: Do the `select`- and `use`- events not need to be tested?

src/Surfnet/Stepup/Tests/Configuration/InstitutionConfigurationTest.php
```diff
@@ -46,6 +49,9 @@ public function use_ra_locations_option_is_set_to_false_by_default_upon_creation
         $verifyEmailOption               = new VerifyEmailOption(true);
         $numberOfTokensPerIdentityOption = new NumberOfTokensPerIdentityOption(0);
         $expectedUseRaLocationsOption = new UseRaLocationsOption(false);
+        $useRaOption = new UseRaOption(null);
+        $useRaaOption = new UseRaaOption(null);
+        $selectRaaOption = new SelectRaaOption(null);
 
         $this->scenario
             ->when(function () use ($institution, $institutionConfigurationId, $showRaaContactInformationOption, $verifyEmailOption) {
```
pablothedude: Is it possible to test in the `then` if the sub events are thrown?

src/Surfnet/Stepup/Tests/Configuration/Event/EventSerializationAndDeserializationTest.php
MKodde: In the context of serialization, they are tested in the very test case you are commenting on. Or are you suggesting another type of test? I'm all ears!

src/Surfnet/Stepup/Tests/Configuration/InstitutionConfigurationTest.php
MKodde: The then describes the expected outcome of, in this case, the creation of the InstitutionConfiguration aggregate. In the annotated test, the 'sub events (`applyUseRaOptionChangedEvent`)' are not triggered.

MKodde: For that to happen we would need to trigger the `rebuild` method on the aggregate. Which would also be an interesting thing to test in the `InstitutionConfigurationTest`.

MKodde: Testing the rebuild method was out of scope for this test, I've added a test to cover the FGA configure options on the aggregate. See commit ea844e6

src/Surfnet/Stepup/Tests/Configuration/Event/EventSerializationAndDeserializationTest.php
pablothedude: I ment I did mis the `Value`-tests for the `Options`. Sorry for mentioning `events`.

src/Surfnet/Stepup/Tests/Configuration/InstitutionConfigurationTest.php
pablothedude: Thanks for looking into this.

src/Surfnet/Stepup/Tests/Configuration/Event/EventSerializationAndDeserializationTest.php
MKodde: Ah, yes, adding a test for value objects representing the FGA options might be a good idea. And I've considered this. Given the very low complexity of said objects. I've restrained from doing this. But now I'm starting to think otherwise.

PR #233, opened 2018-09-18
Update projections for the FGA settings

MKodde:
**What's new?**
This PR is built on top of #232 before merging, make sure that other PR is approved. And rebase any changes onto this branch.

1. The institution configuration options projection is updated, the FGA options are now part of that projection. 
2. The aggregate was updated once again, as nulling the FGA options did not work correctly yet.
3. A new migration was added to this PR that stores the FGA options on the Institution configuration options entity. This is a first place to store this information. A new entity might be introduced that will be dedicated to the FGA options and make them easily retrievable if the need arises.

**What's left to do?**
See the tasks on the ticket for more info on this topic.

**How to test?**
See #232

See https://www.pivotaltracker.com/story/show/160546654

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,3 +36,4 @@ branches:
   only:
     - master
     - develop
+    - feature/update-projections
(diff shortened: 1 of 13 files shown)


MKodde:
I'm going to merge this back to #232 now.

PR #238, opened 2018-10-04
Fix removal of an Institution from the whitelist

pablothedude:
At the moment it's impossible to remove a Institution from the whitelist
if this is done with the endpoint to remove a singel Institution. This
commit will make sure that this will be possible.

https://www.pivotaltracker.com/story/show/160959676


pablothedude:
Already in develop

PR #261, opened 2019-01-25
Fix authorization filter

pablothedude:
In the filtering based on the authorization, there is a bug which prevents fetching the correct RA candidates and RAs themselves.

The problem was that initially the flattened ra_lsiting and ra_candidate listings didn't store an identity in combination with a ra_institution and joined the identity based on the institution configuration. We later decided to flatten the projections and store those entries with both identity and raInstitution. Therefore the join wasn't valid anymore and should have been removed.

Also, there seemed to be a problem with the reuse of the ra_lising projections to determine authorizations based on the view which should be fixed later on by for example introduce a new endpoint in MW for authorizations. This to prevent breaking things when changing unrelated logic. This could be better addressed when solving #163520641 (remove RAA switcher).

I also encountered that in the RA environment the select box to select an institution when selecting an institution for an identity didn't work as expected. This should also be addressed when fixing the RAA switcher. I've also created an issue for this #163520412

## Done ##
https://www.pivotaltracker.com/story/show/163520315
https://www.pivotaltracker.com/story/show/163520412
https://www.pivotaltracker.com/story/show/163520641

### DIFF
--- a/src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/RaCandidateRepository.php
+++ b/src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/RaCandidateRepository.php
@@ -163,9 +163,13 @@ class RaCandidateRepository extends EntityRepository
         $queryBuilder = $this->createQueryBuilder('rac');
 
         // Modify query to filter on authorization
-        $queryBuilder
-            ->andWhere('rac.raInstitution = :raInstitution')
-            ->setParameter('raInstitution', $query->authorizationContext->getActorInstitution());
+        $this->authorizationRepositoryFilter->filterCandidate(
+            $queryBuilder,
+            $query->authorizationContext->getActorInstitution(),
+            'rac.identityId',
+            'rac.raInstitution',
+            'iac'
+        );
 
         if ($query->institution) {
             $queryBuilder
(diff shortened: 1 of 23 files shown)


src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/RaCandidateRepository.php
```diff
@@ -163,13 +163,9 @@ public function createSearchQuery(RaCandidateQuery $query)
...
 
         // Modify query to filter on authorization
-        $this->authorizationRepositoryFilter->filter($queryBuilder, $query->authorizationContext, 'rac.identityId', 'rac.institution', 'iac');
-
-        if ($query->actorInstitution) {
-            $queryBuilder
-                ->andWhere('rac.raInstitution = :raInstitution')
-                ->setParameter('raInstitution', $query->actorInstitution);
-        }
+        $queryBuilder
```
MKodde: The authorizationRepositoryFilter is no longer in use in the RaCandidateRepository, but you explicitly changed the `InstitutionAuthorizationRepositoryFilter` in this commit. I assume to fix an issue in the IdentityRepository or in one of the other repos created by the `AuthorizationRepositoryFactory`? I don't see cohesion yet.

PR #267, opened 2019-02-14
Return highest ra(a) level from ra credentials endpoint

pablothedude:
In MW when requesting the RA-credentials to login in for example RA the first RA-entry was returned. This prevented users which were RAA for another institution to determine the highest RAA level. This is used in for RA to determin if a user should be granted access to a certain route.
In MW the actorId makes sure only entries for the institutions the user is RA(A) for should be returned.


pablothedude:
https://www.pivotaltracker.com/n/projects/1163646/stories/163692589

MKodde:
Great change! One nitpick: could you update the PR title and description to better fit the actual changes made?

pablothedude:
@MKodde: updated the title and description

PR #268, opened 2019-02-25
Fix ra-listing static factory method

pablothedude:
After the fix to be able to login with the highest ra(a) level the
original static method was removed, but this shouldn't be removed.
Restored the original method to allow construction of ra-listings.


pmeulen:
This bug was triggerd in the /registration/yubikey/prove-possession middleware command. Could you add a unit test that exercises this code path?

pablothedude:
I will create a new ticket to add the required test.

MKodde:
It might not be nessecary to create a ticket, these tests already exist.

And your response to point 3?

pablothedude:
The latter

MKodde:
Please explain:

> The latter

PR #269, opened 2019-02-28
Include implicit roles on the profile endpoint

MKodde:
By including the implicit roles on the profile page, the user can see for which institutions he/she can perform RA(A) tasks.

This PR might be a little rough around the edges. @pablothedude feel free to rework this PR. I would also greatly appreciate your feedback on my solution. Your proposal did not seem to cut the bill fully, but I might have misunderstood some of your intentions.

This is in support of https://www.pivotaltracker.com/story/show/164244620 (task 2)
And behat test coverage is found in: https://github.com/OpenConext/Stepup-Deploy/pull/83

### DIFF
--- a/src/Surfnet/StepupMiddleware/ApiBundle/Controller/ProfileController.php
+++ b/src/Surfnet/StepupMiddleware/ApiBundle/Controller/ProfileController.php
@@ -18,8 +18,6 @@
 
 namespace Surfnet\StepupMiddleware\ApiBundle\Controller;
 
-use Surfnet\Stepup\Configuration\Value\InstitutionRole;
-use Surfnet\StepupMiddleware\ApiBundle\Authorization\Value\InstitutionRoleSet;
 use Surfnet\StepupMiddleware\ApiBundle\Identity\Service\ProfileService;
 use Symfony\Bundle\FrameworkBundle\Controller\Controller;
 use Symfony\Component\HttpFoundation\JsonResponse;
@@ -34,17 +32,10 @@ class ProfileController extends Controller
      */
     private $profileService;
 
-    /**
-     * @var InstitutionRoleSet
-     */
-    private $roleRequirements;
-
-    public function __construct(ProfileService $profileService)
-    {
+    public function __construct(
+        ProfileService $profileService
+    ) {
         $this->profileService = $profileService;
-        $this->roleRequirements = new InstitutionRoleSet(
-            [new InstitutionRole(InstitutionRole::ROLE_USE_RA), new InstitutionRole(InstitutionRole::ROLE_USE_RAA)]
-        );
     }
 
     public function getAction(Request $request, $identityId)
--- a/src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/InstitutionListingRepository.php
+++ b/src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/InstitutionListingRepository.php
@@ -105,7 +105,7 @@ class InstitutionListingRepository extends EntityRepository
                 InstitutionAuthorization::class,
                 'a',
                 Join::WITH,
-                "i.institution = a.institution AND a.institutionRole IN (:authorizationRoles)"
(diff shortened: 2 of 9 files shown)


src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/InstitutionListingRepository.php
```diff
@@ -105,7 +105,7 @@ public function getInstitutionsForSelectRaa(IdentityId $actorId)
                 InstitutionAuthorization::class,
                 'a',
                 Join::WITH,
-                "i.institution = a.institution AND a.institutionRole IN (:authorizationRoles)"
+                "r.institution = a.institution AND a.institutionRole IN (:authorizationRoles)"
```
MKodde: In the commit message, you state that there is a bug in this query. The bug is that the join was on the wrong table when looking at the code. 

Is my conclusion correct that: We now only return institutions that actually have a RA? As you now join on the RA Listing instead of joining on the Institution listing.

src/Surfnet/StepupMiddleware/ApiBundle/Identity/Service/ProfileService.php
```diff
@@ -68,53 +75,26 @@ public function __construct(
...
-
-    /**
-     * @param IdentityId $identity
-     * @return AuthorizedInstitutionCollection
-     */
-    private function findAuthorizationsBy(IdentityId $identity)
-    {
-        $ra = new InstitutionRole(InstitutionRole::ROLE_USE_RA);
-        $raa = new InstitutionRole(InstitutionRole::ROLE_USE_RAA);
+        $authorizationContextRa = $this->authorizationService->buildInstitutionAuthorizationContext(
```
MKodde: I like how much the code cleaned up with your change!

MKodde:
The now dead code from https://github.com/OpenConext/Stepup-Middleware/pull/269/commits/36ebde93b90dc656595023bce6868f509b90da7a can be cleaned up. You can probably just squash that commit and its predecessor? This is more elegant than reverting some code changes I made with your own in my opinion.

src/Surfnet/StepupMiddleware/ApiBundle/Controller/ProfileController.php
```diff
@@ -34,17 +32,10 @@ class ProfileController extends Controller
...
     private $profileService;
 
-    /**
-     * @var InstitutionRoleSet
-     */
-    private $roleRequirements;
-
-    public function __construct(ProfileService $profileService)
-    {
+    public function __construct(
```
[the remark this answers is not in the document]
pablothedude: True

src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/InstitutionListingRepository.php
```diff
@@ -206,16 +164,34 @@ public function getInstitutionsForSelectRaaAsSraa(Institution $institution)
...
+    }
+
+    /**
+     * @param InstitutionRole $role
+     * @return array
+     */
+    private function getAuthorizationRolesForRa(InstitutionRole $role)
+    {
+        switch (true) {
+            case $role->equals(InstitutionRole::useRa()):
```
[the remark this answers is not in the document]
pablothedude: No it's used for the join on the roles. If you need the RA-role we should fetch roles which are RA or RAA and not only RA's. If you want want the RAA role you only want roles which are RAA.

src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/InstitutionListingRepository.php
pablothedude: Your conclusion is correct if `:roles` = 'ra'

src/Surfnet/StepupMiddleware/ApiBundle/Identity/Service/ProfileService.php
pablothedude: :+1:

PR #270, opened 2019-03-05
Add filter options to pagination results

pablothedude:
In order to use the filters on the ra-second-factors ra-listing and
ra-candidate page we need to return the possible available filters.

### DIFF
--- a/src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/InstitutionListingRepository.php
+++ b/src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/InstitutionListingRepository.php
@@ -142,7 +142,7 @@ class InstitutionListingRepository extends EntityRepository
                 InstitutionAuthorization::class,
                 'a',
                 Join::WITH,
-                "r.institution = a.institution AND a.institutionRole IN (:authorizationRoles)"
+                "i.institution = a.institution AND a.institutionRole IN (:authorizationRoles)"
             )
             ->where("a.institution = :institution")
             ->groupBy("a.institution");
--- a/src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/RaCandidateRepository.php
+++ b/src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/RaCandidateRepository.php
@@ -202,6 +202,27 @@ class RaCandidateRepository extends EntityRepository
     }
 
     /**
+     * @param RaCandidateQuery $query
+     * @return \Doctrine\ORM\Query
+     */
+    public function createOptionsQuery(RaCandidateQuery $query)
+    {
+        $queryBuilder = $this->createQueryBuilder('rac')
+            ->select('rac.institution')
+            ->groupBy('rac.institution');
+
+        // Modify query to filter on authorization
+        $this->authorizationRepositoryFilter->filter(
+            $queryBuilder,
+            $query->authorizationContext,
+            'rac.raInstitution',
+            'iac'
+        );
+
+        return $queryBuilder->getQuery();
+    }
+
+    /**
      * @param string[] $sraaList
      * @return RaCandidate[]
      */
--- a/src/Surfnet/StepupMiddleware/ApiBundle/Identity/Service/RaCandidateService.php
+++ b/src/Surfnet/StepupMiddleware/ApiBundle/Identity/Service/RaCandidateService.php
@@ -51,6 +51,26 @@ class RaCandidateService extends AbstractSearchService
     }
 
     /**
+     * @param RaCandidateQuery $query
+     * @return array
+     */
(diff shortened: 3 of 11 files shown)


src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/InstitutionListingRepository.php
```diff
@@ -142,7 +142,7 @@ public function getInstitutionsForSelectRaaAsSraa(Institution $institution)
                 InstitutionAuthorization::class,
                 'a',
                 Join::WITH,
-                "r.institution = a.institution AND a.institutionRole IN (:authorizationRoles)"
+                "i.institution = a.institution AND a.institutionRole IN (:authorizationRoles)"
```
MKodde: Didn't you bugfix this yesterday? Was that fix not correct? Or am I missing something here?

src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/RaCandidateRepository.php
```diff
@@ -201,6 +201,27 @@ public function createSearchQuery(RaCandidateQuery $query)
         return $queryBuilder->getQuery();
     }
 
+    /**
+     * @param RaCandidateQuery $query
+     * @return \Doctrine\ORM\Query
+     */
+    public function createOptionsQuery(RaCandidateQuery $query)
```
MKodde: For future maintainability, it might be a good idea to move this logic (that is duplicated in the other repositories) to a helper class that is injected into those repos. This prevents code duplication and improves maintainability. But only if this does not raise other problems.

src/Surfnet/StepupMiddleware/ApiBundle/Identity/Service/RaCandidateService.php
```diff
@@ -50,6 +50,26 @@ public function search(RaCandidateQuery $query)
         return $paginator;
     }
 
+    /**
+     * @param RaCandidateQuery $query
+     * @return array
+     */
+    public function getFilterOptions(RaCandidateQuery $query)
```
MKodde: Like my previous comment, please consider moving this logic to a helper class.

src/Surfnet/StepupMiddleware/ApiBundle/Response/JsonCollectionResponse.php
```diff
@@ -51,6 +54,10 @@ public function __construct($totalItems, $page, $pageSize, array $collection, $h
             'items'      => $collection
         );
 
+        if (!empty($filters)) {
+            $data['filters'] = $filters;
```
MKodde: filters might not be set on the response data. This might cause problems in RA.

src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/InstitutionListingRepository.php
pablothedude: Yes, I thought I needed to fix this but this is not going to work for the query for SRAA's because we don;t have an alias `r`.

src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/RaCandidateRepository.php
pablothedude: I think for now I find the readability better than to split the minimal code duplication. The required (number of) columns do also have some differences.

src/Surfnet/StepupMiddleware/ApiBundle/Identity/Service/RaCandidateService.php
pablothedude: See my comment above

src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/InstitutionListingRepository.php
MKodde: Acknowledged!

src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/RaCandidateRepository.php
MKodde: Agreed!

src/Surfnet/StepupMiddleware/ApiBundle/Response/JsonCollectionResponse.php
MKodde: Agreed, please don't forget the update of the API docs!

pablothedude: Fixed this to always set data

PR #272, opened 2019-03-05
Fix ra second factor auth level

pablothedude:
The authorization level was set to RAA while RA should be already valid.


pablothedude:
https://www.pivotaltracker.com/story/show/164398880

PR #273, opened 2019-03-05
Fix invalid left join for authorizations

pablothedude:
The left join needed to be removed to prevent null values for
institutions. While fetching the authorized roles.


pablothedude:
Merged back with https://github.com/OpenConext/Stepup-Middleware/pull/274

PR #274, opened 2019-03-06
Add management institutions to the profile endpoint

pablothedude:
This is needed in order to get the management rights through select_raa.


src/Surfnet/StepupMiddleware/ApiBundle/Identity/Value/Profile.php
```diff
@@ -33,6 +34,11 @@ class Profile implements JsonSerializable
      */
     private $authorizedInstitutionCollection;
 
+    /**
+     * @var InstitutionCollection
+     */
+    private $selectRaa;
```
MKodde: I think this field would benefit from some additional information. We all know the select_raa institution configuration option, but what does it mean in the context of the profile? 

If I interpret this correctly, these are the implicit RAA roles that apply for the user right? Maybe using a term along those lines would be more expressive? Or like I requested, add an explanation.

PR #276, opened 2019-04-24
Add additional allowed sandbox methods

MKodde:
The implicit __toString call is no longer allowed, and a critical error is raised by Twig. The solution is to add the __toString calls to the list of allowed sandbox methods.

This change fixes the problem where sending the verification mail message would trigger an error.

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -45,6 +45,8 @@ services:
             - [ if, else, elseif, for ] # Allowed tags
             - [ escape, localizeddate ] # Allowed filters
             - # Allowed methods
+                Surfnet\Stepup\Identity\Value\CommonName:
+                    - __toString
                 Surfnet\Stepup\Configuration\Value\ContactInformation:
                     - __toString
                 Surfnet\Stepup\Configuration\Value\Location:
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -45,6 +45,14 @@ services:
             - [ if, else, elseif, for ] # Allowed tags
             - [ escape, localizeddate ] # Allowed filters
             - # Allowed methods
+                Surfnet\Stepup\Configuration\Value\ContactInformation:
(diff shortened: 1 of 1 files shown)


MKodde:
Note that the build fails on the Symfony Bridge CVE, which will be fixed in #277

pablothedude:
If in the meantime the table content is changed and someone started to use RaLocation::$institution this could stop working. Should `\Surfnet\Stepup\Configuration\Value\Institution::__toString` also be added to the configuration array?

MKodde:
@pmeulen, The consequences of the fix to the recent Twig Sandbox vulnerability have been addressed in this PR. As you know, the mail templates are configured using the MW configuration API. And the Twig templates in that config are parsed using Twig in sandbox mode. 

The TLDR of the vulnerability was, that Twig in sandbox mode automatically tried to use the __toString magic method when outputting a variable that had not been whitelisted in the sandbox config.

My fix was to add the implicitly called __toString method calls to the whitelist in the twig sandbox configuration.

Question. Have the mail templates further diverged, are they significantly different from the ones found in Stepup-Deploy? If they are, I might have to add additional entries on the whitelist, like @pablothedude already suggested.

MKodde:
>  to

In the previous situation, that would also have required a config change. As we are only allowed to access the name, location and contact information from the RALocation entity. I did add the CommonName VO to the list of allowed methods.

pmeulen:
> Question. Have the mail templates further diverged, are they significantly different from the ones found in Stepup-Deploy? If they are, I might have to add additional entries on the whitelist, like @pablothedude already suggested.

The mail templates in https://github.com/OpenConext/Stepup-Deploy/tree/develop/environments/template/templates/middleware are representative of what SURFsecureID uses in production. The differences are textual.

MKodde:
Thanks @pmeulen. Merging

PR #279, opened 2019-05-02
Add Self Service verified SF collection endpoint

MKodde:
This endpoint does not apply the FGA authz filtering. Allowing the self
service user to see all it's tokens regardless of institution

### DIFF
--- a/src/Surfnet/StepupMiddleware/ApiBundle/Controller/VerifiedSecondFactorController.php
+++ b/src/Surfnet/StepupMiddleware/ApiBundle/Controller/VerifiedSecondFactorController.php
@@ -65,7 +65,7 @@ class VerifiedSecondFactorController extends Controller
 
     public function collectionAction(Request $request)
     {
-        $this->denyAccessUnlessGranted(['ROLE_RA', 'ROLE_SS']);
+        $this->denyAccessUnlessGranted(['ROLE_RA']);
 
         $actorId = new IdentityId($request->get('actorId'));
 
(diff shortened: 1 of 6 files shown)


pablothedude:
The changes themselves are looking good. I would only suggest to name the endpoint to inform it's used for listing tokens from a identity itself. So we could lose not used logic for this specific endpoint to lower additional complexity.

src/Surfnet/StepupMiddleware/ApiBundle/Controller/VerifiedSecondFactorController.php
```diff
@@ -65,10 +65,32 @@ public function getAction($id)
...
+            $actorId,
+            new InstitutionRole(InstitutionRole::ROLE_USE_RA)
+        );
+
+        $paginator = $this->secondFactorService->searchVerifiedSecondFactors($query);
+
+        return JsonCollectionResponse::fromPaginator($paginator);
+    }
+
+    public function collectionWithoutAuthorizationContextAction(Request $request)
```
pablothedude: I think this could be better renamed to be more descriptive for example `tokensOfIdentityCokkectionAction` or something like that. Because it can only be used by SS and is only used to list own tokens we should drop complexity if we can.

src/Surfnet/StepupMiddleware/ApiBundle/Controller/VerifiedSecondFactorController.php
```diff
@@ -65,10 +65,32 @@ public function getAction($id)
...
+    {
+        $this->denyAccessUnlessGranted(['ROLE_SS']);
+        $query = $this->buildVerifiedSecondFactorQueryFrom($request);
```
pablothedude: Why not only set `identityId`?

src/Surfnet/StepupMiddleware/ApiBundle/Resources/config/routing.yml
```diff
@@ -38,6 +38,12 @@ verified_second_factors:
     methods:  [GET]
     condition: "request.headers.get('Accept') matches '/^application\\\\/json($|[;,])/'"
 
+verified_second_factors_no_authz:
+    path:     /self-service-verified-second-factors
```
pablothedude: Please rename to be more descriptive from the viewpoint of SS and current identity.

MKodde: Yes, renamed to `verified-second-factors-of-identity`

src/Surfnet/StepupMiddleware/ApiBundle/Controller/VerifiedSecondFactorController.php
MKodde: Good suggestion, refactored the two collection methods.

src/Surfnet/StepupMiddleware/ApiBundle/Controller/VerifiedSecondFactorController.php
MKodde: Renamed to `collectionOfIdentityAction`

pablothedude:
Thanks for the changes, this will makes this a lot less complex and they will thank you for that in the future!

PR #281, opened 2019-05-10
Add exception for not whitelisted RA's

pablothedude:
In order to clarify the reason why a RA which institution isn't on the
whitelist isn't allowed this exception is added. Before this resulted
in an exception thrown when trying to access a method on a null object.

`Error: Call to a member function isSraa() on null`


pablothedude:
https://www.pivotaltracker.com/story/show/165850321

PR #286, opened 2019-11-07
Upgrade Stepup-saml-bundle to version 4.1.8

MKodde:
This change will apply the countermeasures to harden against
CVE 2019-3465 and will effectively bump `robrichards/xmlseclibs` to
version 3.0.4

This release is targetted at the R17 (3.0.x) code wich currently resides on develop/master

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 3.0.1
+This is a security release that will harden the application against CVE 2019-346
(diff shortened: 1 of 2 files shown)


pablothedude:
Are all updated dependencies needed for the security update. I certainly don't mind if they are getting updated now as the changes would land in dev. But I think it would be nice to elaborate why this was done?

MKodde:
Ah, yes. This is the result of a composer update --with-dependencies statement. I don't think this will cause any harm. And for develop -> 3.0.x release. This will be no biggie in my opinion

PR #291, opened 2020-01-16
Fix sho mixed casing once and for all

pablothedude:
The domain accepted mixed casing sho's. This shouldn't be possible.
Therefore this is now enforced in the domain itself and a migration
is added to convert all occurrences to lowercase.

### DIFF
--- a/src/Surfnet/Stepup/Tests/Configuration/Value/InstitutionConfigurationIdTest.php
+++ b/src/Surfnet/Stepup/Tests/Configuration/Value/InstitutionConfigurationIdTest.php
@@ -100,7 +100,7 @@ class InstitutionConfigurationIdTest extends TestCase
      * @test
      * @group domain
      */
-    public function normalized_institution_configuration_ids_and_unnormalized_institution_configuration_ids_are_not_the_same()
+    public function normalized_institution_configuration_ids_and_unnormalized_institution_configuration_ids_are_the_same()
     {
         $mixedCaseInstitution = new Institution('An InStItUtIoN');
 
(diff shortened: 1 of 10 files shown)


MKodde:
Fixes: https://www.pivotaltracker.com/story/show/170487492

src/Surfnet/Stepup/Tests/Configuration/Value/InstitutionConfigurationIdTest.php
```diff
@@ -109,10 +109,8 @@ public function normalized_institution_configuration_ids_and_unnormalized_instit
 
         $isSameId = $unnormalizedInstitutionConfigurationId->equals($normalizedInstitutionConfigurationId);
 
-        $this->assertFalse(
-            $isSameId,
-            'An normalized InstitutionConfigurationId based on an institution with mixed casing'
-            . 'should not match an unnormalized InstitutionConfigurationId based on the same institution'
+        $this->assertTrue(
+            $isSameId
```
MKodde: This can be a one-liner right?

pablothedude:
As far as I could see this seemed not that intentional. The normalization code seemed a fix because earlier on a UUIDv5 was used (hashed institution name based on an unnormalized name) which resulted in multiple UUID's for the same institution configuration and you wouldn't be able to convert this back very easily.

I've looked in the code and I found documentation in the code that stated that the old (unnormalized) logic could be removed in the next release. This would also prevent bugs in the future because the domain forces the use of a normalized institution name throughout the project.

See https://github.com/OpenConext/Stepup-Middleware/commit/c2c68c6650290d7e6f354549cf4b3d49bc9e6495
`@deprecated To be removed in next release; use normalizedFrom method to account for case-(in)sensitivity issues`

I want to open a TD ticket to clean this up later on. For now I think this will be suffiecient and otherwise to big for R17. I will discuss this further IRL.

src/Surfnet/Stepup/Tests/Configuration/Value/InstitutionConfigurationIdTest.php
pablothedude: Done

MKodde:
Yes, let's go ahead with this. Be sure to create a ticket or other means to not forget to address the cleanup task

PR #294, opened 2020-02-03
Add remote vetting command

pablothedude:
A command is added to allow remote vetting of a second factor in order
to be able to deliver a RV POC.


phavekes:
@pmeulen Can this be closed?

PR #296, opened 2020-02-18
Allow language switching from SelfService

MKodde:
The MW command authorization service did pass this command as a valid self service command, but did not pass the successive RA check.

https://www.pivotaltracker.com/story/show/171325732


MKodde:
Good feedback, I totally agree that leaving complexity high (in this case) increases readability of the class' responsibility.

FYI also updated the changelog entry while at it.

PR #297, opened 2020-02-27
Add missing institution to the filter options

MKodde:
The ra-candidate search form allows filtering search results on
institution. But that filter select list was empty. Because the
institutions where not added to the search results.

That is fixed in the RaCandidateRepository::createOptionsQuery method by
also selecting the institutions.

https://www.pivotaltracker.com/story/show/171512620

### DIFF
--- a/src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/RaCandidateRepository.php
+++ b/src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/RaCandidateRepository.php
@@ -216,8 +216,7 @@ class RaCandidateRepository extends EntityRepository
     public function createOptionsQuery(RaCandidateQuery $query)
     {
         $queryBuilder = $this->createQueryBuilder('rac')
-            ->select('rac.raInstitution')
-            ->groupBy('rac.raInstitution');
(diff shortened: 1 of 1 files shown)


src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/RaCandidateRepository.php
```diff
@@ -215,9 +215,9 @@ public function createSearchQuery(RaCandidateQuery $query)
      */
     public function createOptionsQuery(RaCandidateQuery $query)
     {
-        $queryBuilder = $this->createQueryBuilder('rac')
-            ->select('rac.raInstitution')
-            ->groupBy('rac.raInstitution');
+        $queryBuilder = $this->createQueryBuilder('r')
+            ->select('r.institution, r.raInstitution')
```
pablothedude: Isn't one of them currently abundant?

MKodde: Yes it is, my rationale was to leave it. But stripping it away might be a better idea. Fixing this.

PR #298, opened 2020-03-09
Fix invalid RA candidate authorization

pablothedude:
The RA-Candidates were filtered on RA institution but that should have
been on institution. Because we only want to use identities from the
institutions we are currently RA for.

### DIFF
--- a/src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/RaCandidateRepository.php
+++ b/src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/RaCandidateRepository.php
@@ -165,11 +165,14 @@ class RaCandidateRepository extends EntityRepository
     {
         $queryBuilder = $this->createQueryBuilder('rac');
 
-        // Modify query to filter on authorization
+        // Modify query to filter on authorization:
+        // For the RA candidates we want the identities that we could make RA. Because we then need to look at the
+        // select_raa's we have to look at the institution of the candidate because that's the institution we could
+        // select RA's from. Hence the 'rac.institution'.
         $this->authorizationRepositoryFilter->filter(
             $queryBuilder,
             $query->authorizationContext,
-            'rac.raInstitution',
+            'rac.institution',
             'iac'
         );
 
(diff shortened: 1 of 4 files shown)


MKodde:
You change updated the authorization filter field from ra-institution to the institution field. Isn't this always the case(for example also in the RA listing repository)? Can you explain why in certain situations we would have to filter the authorization filter using RA institution, and in other situations use the institution of the user. I thought the authorization filter was designed only to enforce the FGA authorizations.

Wouldn't adding a regular where clause on the institution not be a better solution?

pablothedude:
For the RA listing we want identities that are already RA. Because we then need to look at the `use_raa`'s we have to look at the RA-institutions because that's the institution the user is RA for and we should use those RA's.

For the RA candidates we want the identities that we could make RA. Because we then need to look at the `select_raa`'s we have to look at the institution of the candidate because that's the institution we could select RA's from.

I don't think adding a regular where will be necessarily better but adding and improving repository tests will be. We then could ultimately refactor things to our liking. But I don't want to make that effort right away.
https://www.pivotaltracker.com/story/show/171622570.

MKodde:
Okay, your explanation makes sense. I do think filtering on the institution outside of the authz filtering is the way to go. As this is not a limitation on the data you are allowed to see, but a filter on the data that should be shown from a business rule standpoint.

__What should be changed in my opinion:__

1. Write inline-documentation on why we filter on institution in some cases, and on raInstitiontion in other cases. This would be valuable information for future reference. At least, I really missed that information while reviewing the code.
2. Change your commit message, and include information like you provided me after my first feedback.
3. Add concrete test suggestions to the ticket. I've already added some idea's and a link to this ticket. The more the better!

pablothedude:
> **What should be changed in my opinion:**
> 
>     1. Write inline-documentation on why we filter on institution in some cases, and on raInstitiontion in other cases. This would be valuable information for future reference. At least, I really missed that information while reviewing the code.
Good suggestion, done!
>     2. Change your commit message, and include information like you provided me after my first feedback.
Good suggestion, done!
>     3. Add concrete test suggestions to the ticket. I've already added some idea's and a link to this ticket. The more the better!
Will do this

PR #299, opened 2020-03-12
Show the correct RA candidates for the virtual institution use case

pablothedude:
When retracting an RA role the identity should have been added to the
candidate projection. This however didn't work as intended and the user
was reverted to his own institution and not the ra-institution.

A test was added in deploy to reproduce and mitigate the bug.
https://github.com/OpenConext/Stepup-Deploy/pull/101/commits/2e5e5bee32014c292a2af8e5b070ca9f6dca012c


pablothedude:
As far as I could see I haven't found any regressions in the presentation of the institution authorization configuration on the profile page.

PR #300, opened 2020-03-12
Disable ra-candidate fulltext search

pablothedude:
Fulltext search didn't result in the results expected. Therefore
the query was therefore changed to used a regular like statement.

https://www.pivotaltracker.com/story/show/171812850


pablothedude:
Do you have any reliable sources for the assumption that match_against is faster than a simple like query? Because a fulltext index has to be  much larger this seems almost impossible and I've never encountered this behaviour myself.

A good suggestion to change the match_against in the IdentityRepository.

MKodde:
> Do you have any reliable sources for the assumption that match_against is faster than a simple like query?

I queried google results regarding this topic. One writer performed tests on different storage engines, combined with using optimized match queries. Match against queries sometimes trigger a full table scan, making them much slower than a like clause on a well-indexed column. Given the simple filtering we try to achieve here, going for LIKE is the best option IMHO.

MKodde:
And now that we no longer utilize the MATCH_AGAINST feature, we can rid ourselves of that dql string_function

https://github.com/OpenConext/Stepup-Middleware/blob/4e1ba1ef933df6d6b8b3940b88f44f1487df1f5d/app/config/config.yml#L159

PR #301, opened 2020-04-16
Remove RaCandidate projector in favor of query

pablothedude:
This is done in order to mitigate performance issues when updating the institution configuration.
If the configuration was updated all ra candidates for select_raa institutions needed to be checked
and updated if required. Which resulted in timouts and OOM's.

Initially a view was used to prevent updateing the RACandidate through events but this proved
slow when filtering was done on the view in some cases. Therefore the view was removed and
the query itself was moved to the repository so filtering could be handled earlier on which
greatly  improved performance.

https://www.pivotaltracker.com/story/show/172245466
https://github.com/OpenConext/Stepup-Deploy/pull/105


README.md
```diff
@@ -33,7 +33,7 @@ Some of the configuratio of the components is static (i.e. stored in parameteres
 
 Whenever adding a new event, be sure to update `app/config/events.yml`.
 This is a list of events that is shown when replaying events.
-Also be sure to create or update the event serialization/deserialization tests,
+Also be sure to create or update the event serialization/deserialization tests,RaCandidateRepository
```
MKodde, line 36: Accidental pasted RaCandidateRepository?

app/DoctrineMigrations/Version20200416135127.php
```diff
@@ -0,0 +1,49 @@
...
+                i.email,
+                i.name_id,
+                a.institution AS ra_institution
+FROM identity i
+         INNER JOIN vetted_second_factor vsf on vsf.identity_id = i.id
+         INNER JOIN institution_authorization a ON (
+        a.institution_role = \'select_raa\' AND a.institution_relation = i.institution
+    )
+WHERE NOT EXISTS(SELECT 1 FROM ra_listing AS l WHERE l.identity_id = i.id AND l.ra_institution = a.institution)
+);');
```
MKodde, line 35: 3 nitpicks:
1. For the sake of readability consider using HEREDOC to capture the query.
2. My inner SQL linter triggered on mixed-case use of MySQL keywords (`AS` and `as`, `on`). You also not use them consistently. For the aliases you create for the table names (sh|c)ould also have an `AS` keyword.
3. Consider changing the where not exists subquery into a multi-liner.

app/DoctrineMigrations/Version20200416135127.php
```diff
@@ -0,0 +1,49 @@
...
+    /**
+     * @param Schema $schema
+     */
+    public function down(Schema $schema)
+    {
+        // this down() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql', 'Migration can only be executed safely on \'mysql\'.');
+
+        $this->addSql('DROP VIEW `view_ra_candidate` ');
+        $this->addSql('DROP TABLE ra_candidate');
```
MKodde, line 47: I'm confused why the ra_candidate table is also dropped when the migration is downed. Would expect it to be dropped on the UP method. Could you explain this maybe also in an in-line comment?

src/Surfnet/StepupMiddleware/ApiBundle/EventListener/IgnoreTablesListener.php
```diff
@@ -0,0 +1,70 @@
...
+
+    public function postGenerateSchema(GenerateSchemaEventArgs $args)
+    {
+        $schema = $args->getSchema();
+
+        $tableNames = $schema->getTableNames();
+        foreach ($tableNames as $tableName) {
+            if (in_array($tableName, $this->ignoredTables)) {
+                // remove table from schema
+                $schema->dropTable($tableName);
```
MKodde, line 45: What a confusing methodName for that action. It implies dropping the view/table from the database schema. Where it actually stops tracking it. Correct me if I'm wrong. And yes, I know this is Doctrine ORM, not our code.

src/Surfnet/StepupMiddleware/ApiBundle/Identity/Entity/RaCandidate.php
```diff
@@ -27,39 +27,24 @@
...
- *          @ORM\Index(name="idxft_ra_candidate_email", columns={"email"}, flags={"FULLTEXT"}),
- *          @ORM\Index(name="idxft_ra_candidate_commonname", columns={"common_name"}, flags={"FULLTEXT"}),
- *          @ORM\Index(name="idx_ra_institution", columns={"ra_institution"})
- *      },
- *     uniqueConstraints={
- *          @ORM\UniqueConstraint(name="idx_ra_candidate_unique_identity_institution", columns={"identity_id", "ra_institution"})
- *     }
- * )
+ * Be aware that this entity is used for the RA Candidate view. This entity shouldn't be used to store any RA candidates.
+ * The IgnoreTablesListener is used to prevent schema changes when the Doctrine Schema tool is used.
```
MKodde, line 31: :+1:

src/Surfnet/StepupMiddleware/ApiBundle/Identity/Entity/RaCandidate.php
```diff
@@ -94,6 +79,10 @@ class RaCandidate implements JsonSerializable
      */
     public $email;
 
+    private function __construct()
+    {
+    }
```
MKodde, line 84: Was adding a private constructor important? And why?

pablothedude: This is to prevent instantiating an Entity which shouldn't be initiated at all because a view is used.

src/Surfnet/StepupMiddleware/ApiBundle/EventListener/IgnoreTablesListener.php
pablothedude: I don't like the listener either but that seemed to be the only solution. Initially the view is added so it needs to be dropped to let the schema tool work with the view.

app/DoctrineMigrations/Version20200416135127.php
pablothedude: Thanks, that indeed shouldn't happen at all.

app/DoctrineMigrations/Version20200416135127.php
pablothedude: Done

app/DoctrineMigrations/Version20200416135127.php
pablothedude: Removed it completely. I'll consult our PM what should happen with it. Maybe we could better clean it the next release.

pablothedude: The RA candidate could be removed so I added the drop statement in the `up` method and added a irreversible exception to the `down` method.

pablothedude:
@MKodde I've removed the view and moved the query used to the RaCandidate repository. This greatly optimized the speed to return a result because the DBMS could optimize the query much better.

The filtering on the ra candidate page is changed a bit because it will show all institutions even if there are no possible ra-candidates. This won't outweigh the benefits of the improved performance because the only drawback would be an empty resultset when selected.

I've done the change in a separate commit so it would become easier to review.

src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/RaCandidateRepository.php
```diff
@@ -67,42 +86,42 @@ public function createSearchQuery(RaCandidateQuery $query)
...
             $queryBuilder
-                ->andWhere('rac.email LIKE :email')
+                ->andWhere('i.email LIKE :email')
                 ->setParameter('email', sprintf('%%%s%%', $query->email));
         }
 
         if (!empty($query->secondFactorTypes)) {
             $queryBuilder
-                ->innerJoin(VettedSecondFactor::class, 'vsf', Join::WITH, 'rac.identityId = vsf.identityId')
+                //->innerJoin(VettedSecondFactor::class, 'vsf', Join::WITH, 'rac.identityId = vsf.identityId')
```
MKodde, line 113: Perhaps permanently remove this?

app/DoctrineMigrations/Version20200416135127.php
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    /**
+     * @param Schema $schema
+     */
+    public function down(Schema $schema)
+    {
+        $this->throwIrreversibleMigrationException('This migration is irreversible and cannot be reverted because it will need a replay on the RACandidateProjector.');
+
+        //$this->addSql('DROP INDEX idx_vetted_second_factor_type ON vetted_second_factor');
```
MKodde, line 28: Remove this permanently? If you want to keep it for future reference, then please add a comment.

src/Surfnet/StepupMiddleware/ApiBundle/Identity/Entity/RaCandidate.php
```diff
@@ -27,39 +27,24 @@
...
- *          @ORM\Index(name="idx_ra_institution", columns={"ra_institution"})
- *      },
- *     uniqueConstraints={
- *          @ORM\UniqueConstraint(name="idx_ra_candidate_unique_identity_institution", columns={"identity_id", "ra_institution"})
- *     }
- * )
+ * Be aware that this entity is used for the RA Candidate view. This entity shouldn't be used to store any RA candidates.
+ * The IgnoreTablesListener is used to prevent schema changes when the Doctrine Schema tool is used.
+ *
+ * @ORM\Table(name="view_ra_candidate")
```
MKodde, line 33: This might need revision after venturing away from the view?

app/DoctrineMigrations/Version20200416135127.php
pablothedude: Removed it completely

MKodde:
I've functionally tested this after removing the view. I did not find any functional regressions. The ra candidate listing seemed to remain identical between commits. :+1:

src/Surfnet/StepupMiddleware/ApiBundle/Identity/Entity/RaCandidate.php
pablothedude: Yes, that's true and should have been removed. I've directly removed the `IgnoreTablesListener` and the creation of the view in Deploy which could then also be removed.

PR #302, opened 2020-04-16
Add bootstrap console commands

MKodde:
Allow bootstrapping of an Identity with an SMS token. Three verification states can be achieved: Unverified, Verified and Vetted.

Example usage:
`app/console middleware:bootstrap:identity-with-sms my:fancy:name-id institution-a.example.com "Jesse James" "email@jesse-james.co.uk" en_GB "+31 (0) 612345678" vetted`

Some open ends:
- Solve the authority id issue (which RA is vetting the token?) Either save the actor id on the event or in the metadata. For now does not have to end up in the audit log, but should be saved.
- ~Review the overall structure of the command, and look into making the code more maintainable, readable and robust~
- Discuss the separation of the identity creation logic to another command

PM @phavekes expressed his preference to keep the logic bundled in the command. I did however move the creation logic to an abstract command. That way other similar bootstrap events can borrow that logic. 

- ~Be 'email verification required' aware~

https://www.pivotaltracker.com/story/show/172429585


src/Surfnet/StepupMiddleware/MiddlewareBundle/Console/Command/BootstrapIdentityWithSmsSecondFactorCommand.php
```diff
@@ -0,0 +1,220 @@
...
+                sprintf(
+                    '<notice>An identity with name ID "%s" from institution "%s" already exists, using that identity</notice>',
+                    $nameId->getNameId(),
+                    $institution->getInstitution()
+                )
+            );
+            $identity = $identityRepository->findOneByNameIdAndInstitution($nameId, $institution);
+        }
+
+        $connection->beginTransaction();
```
pablothedude, line 110: Why are you using nested transactions when using the transaction aware pipeline. Are  you sure this is safe to use?

src/Surfnet/StepupMiddleware/MiddlewareBundle/Console/Command/BootstrapIdentityWithSmsSecondFactorCommand.php
```diff
@@ -0,0 +1,220 @@
...
+
+        $secondFactorId = Uuid::uuid4()->toString();
+
+        if (!$identity) {
```
pablothedude, line 114: I think it will be better to remove this, because you should bootstrap an identity and not update an identity so it will be better to throw an exception.

pablothedude:
During testing it seems that adding one user it will take approx 1 second. So if I want to use this to test 1000's of users it will take a huge amount of time. Again another drawback of the chosen MW architecture.

src/Surfnet/StepupMiddleware/MiddlewareBundle/Console/Command/BootstrapIdentityWithSmsSecondFactorCommand.php
MKodde: My work is based on the tried and tested yubikey bootstrap script. I do not see nested transactions at work here. At least not according to this definition:

https://www.doctrine-project.org/projects/doctrine-dbal/en/2.10/reference/transactions.html#transaction-nesting

Can you elaborate?

> Are you sure this is safe to use?

The tests I performed worked out fine, whenever illegal operations where applied, the rollback feature worked a charm.

src/Surfnet/StepupMiddleware/MiddlewareBundle/Console/Command/BootstrapIdentityWithSmsSecondFactorCommand.php
MKodde: You may be right, but at the other hand, it would be rather tedious to script up a system to first create/activate a user, and then register a token. 

I'm asking PMs what their preference is. From a single responsibility standpoint I fully agree with you by the way.

MKodde:
> During testing it seems that adding one user will take approx. 1 to 2 seconds. So if I want to use this to test 1000's of users it will take a huge amount of time. Again another drawback of the chosen MW architecture.

This could be speed up by directly writing to the projections table (and insert the events manually) But this would increase chances to invalidate the event stream even more. As you already discussed with the PM's, running this command without debug mode will greatly speed up the time it takes to run the cmd. Let's see if this is acceptable.

> Also SMS is not a LOA3 token so it is not possible to test the RaCandidates with this.

I'm aware of this. That does not mean we do not want to be able to bootstrap SMS tokens. The goal was to be able to add all types of tokens. And SMS was the first one I attempted.

> This also isn't going to work for institution-b because then the token doesn't have to get verified.

Yes, I'm going to have to look to the mail verification institution flag in order to require or skip mail verification. That was on my todo list, thanks for noting it!

src/Surfnet/StepupMiddleware/MiddlewareBundle/Console/Command/BootstrapIdentityWithSmsSecondFactorCommand.php
pablothedude: In the TransactionAwarePipeline multiple connections are used and separate transactions are used. So you need to be sure that if a rollback is performed this will happening on all connections. Personally I think it's not safe to assume the old code is very well battle tested based on earlier discovered bugs.

src/Surfnet/StepupMiddleware/MiddlewareBundle/Console/Command/BootstrapIdentityWithSmsSecondFactorCommand.php
pablothedude: I don't mean to do either one of them (create identity or vet a token) but always do both, so please remove this check and always create an identity with a token. So if the Identity already exists throw an exception.

src/Surfnet/StepupMiddleware/MiddlewareBundle/Console/Command/AbstractBootstrapCommand.php
```diff
@@ -0,0 +1,93 @@
...
+    /** @var IdentityRepository  */
+    protected $identityRepository;
+    /** @var UnverifiedSecondFactorRepository  */
+    protected $unverifiedSecondFactorRepository;
+    /** @var VerifiedSecondFactorRepository */
+    protected $verifiedSecondFactorRepository;
+    /** @var TokenStorageInterface */
+    protected $tokenStorage;
+
+    public function __construct(
```
pablothedude, line 58: I don't like the abstract constructor because if something is likely to change it will be the constructor. The abstract generation of commands or identities could indeed benefit from this setup so you could leave them but only return a command which then could be injected into the commandbus by the concrete implementations themselves.

src/Surfnet/StepupMiddleware/MiddlewareBundle/Resources/config/console_commands.yml
```diff
@@ -0,0 +1,17 @@
+services:
+    Surfnet\StepupMiddleware\MiddlewareBundle\Console\Command\AbstractBootstrapCommand:
```
pablothedude, line 2: Please don't use an abstract constructor

pablothedude:
Functional testing:
- Travis needs to be fixed
- It is possible to add tokens on a already persistent identity, so it will not only bootstrap an identity.

src/Surfnet/StepupMiddleware/MiddlewareBundle/Console/Command/BootstrapIdentityWithSmsSecondFactorCommand.php
MKodde: The benefit of this system is that you can also create multi-token users this way. Something that is interesting from a test perspective.

MKodde: @pmeulen and/or @phavekes what's your preference?

MKodde:
@pablothedude I'd like to propose to keep this feature branch as a platfrom to create the other bootstrap commands. Is that OK with you?

MKodde:
My first branch will be to decouple the sms bootstrap command from the creation of an identity.

src/Surfnet/StepupMiddleware/MiddlewareBundle/Console/Command/BootstrapIdentityWithSmsSecondFactorCommand.php
MKodde: Discussed this. We decided upon a solution where we have separate commands for identity and token cration.

src/Surfnet/StepupMiddleware/MiddlewareBundle/Resources/config/console_commands.yml
MKodde: For now I lean towards leaving it as is. I know its not common practice to use a shared constructor. But given the fact they actually all use the same dependencies. This is something I can get away with (imho). Once dependencies start to differ, this should definitely be reverted.

src/Surfnet/StepupMiddleware/MiddlewareBundle/Console/Command/AbstractBootstrapCommand.php
MKodde: See comment below

MKodde:
@pablothedude can we go forward with this?

pablothedude:
Feel free to do so, it's not even production code, so it would be easy to do some tweaks later on the road if we see fit. It's already a great addition to have these commands to bootstrap large batches.

PR #304, opened 2020-04-21
Bootstrap yubikey command

MKodde:
This PR adds the possibility to bootstrap a Yubikey token

In the process a method was moved to the abstract command. This was also trickled down to the SMS bootstrap command.

https://www.pivotaltracker.com/story/show/172429585


src/Surfnet/StepupMiddleware/MiddlewareBundle/Console/Command/BootstrapYubikeySecondFactorCommand.php
```diff
@@ -0,0 +1,148 @@
...
+
+            return;
+        }
+        $identity = $this->tokenBootstrapService->findOneByNameIdAndInstitution($nameId, $institution);
+        $output->writeln(sprintf('<comment>Adding a %s Yubikey token for %s</comment>', $registrationStatus, $identity->commonName));
+        $this->beginTransaction();
+        $secondFactorId = Uuid::uuid4()->toString();
+
+        try {
+            switch ($registrationStatus) {
```
pablothedude, line 83: I think it would be valuable to add a `default` with an exception, to catch wrong statuses that otherwise get unnoticed.

MKodde: Yes that wouild be a good idea, will open a PR to address this

PR #308, opened 2020-06-07
Allow setting encryption and alternative mail port

tvdijen:


app/config/parameters.yml.dist
```diff
@@ -56,4 +58,4 @@ parameters:
     # This is the global, application wide default. This configuration should be specified for each institution in the
     # institution config (middleware api). The value configured in the parameters.yml will be used as the
     # fallback/default value.
-    number_of_tokens_per_identity: 1
\ No newline at end of file
+    number_of_tokens_per_identity: 1
```
MKodde, line 61: :muscle:

MKodde (approved):
Thanks Tim for expanding the mail settings. I see no issue with this. Maybe setting the mailer port to 25 (which is the default IIRC) is not really required? A `~` might be sufficient here?

PR #309, opened 2020-07-17
Make prove possession step optional

pablothedude:
**Be aware that this is build on top of #307  so this needs to be merged back first!**

This change will effectively make the the prove possession step optional. It's possible to configure the second factors which do not require that step.

- [x] Changelog documentation #310 

https://www.pivotaltracker.com/story/show/172911922

### DIFF
--- a/src/Surfnet/Stepup/Helper/SecondFactorProvePossessionHelper.php
+++ b/src/Surfnet/Stepup/Helper/SecondFactorProvePossessionHelper.php
@@ -1,0 +1,56 @@
+<?php
+/**
+ * Copyright 2020 SURFnet B.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Surfnet\Stepup\Helper;
+
+use Surfnet\StepupBundle\Service\SecondFactorTypeService;
+use Surfnet\StepupBundle\Value\SecondFactorType;
+use Webmozart\Assert\Assert;
+
+class SecondFactorProvePossessionHelper
+{
+    /**
+     * @var array
+     */
+    private $skipProvePossessionSecondFactorTypes;
+
+    /**
+     * @param SecondFactorTypeService $secondFactorTypeService
+     * @param array $skipProvePossessionSecondFactorTypes
+     */
+    public function __construct(
+        SecondFactorTypeService $secondFactorTypeService,
+        array $skipProvePossessionSecondFactorTypes
+    ) {
+        Assert::allInArray(
+            $skipProvePossessionSecondFactorTypes,
+            $secondFactorTypeService->getAvailableSecondFactorTypes(),
+            'Unsupported second factor type configured to skip prove possession'
+        );
+
+        $this->skipProvePossessionSecondFactorTypes = $skipProvePossessionSecondFactorTypes;
+    }
+
+    /**
+     * @param SecondFactorType $secondFactorType
+     * @return bool
+     */
+    public function canSkipProvePossession(SecondFactorType $secondFactorType)
+    {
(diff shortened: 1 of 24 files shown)


src/Surfnet/Stepup/Helper/SecondFactorProvePossessionHelper.php
```diff
@@ -0,0 +1,60 @@
...
+     * @param SecondFactorType $secondFactorType
+     * @return bool
+     */
+    public function canSkipProvePossession(SecondFactorType $secondFactorType)
+    {
+        if (in_array($secondFactorType->getSecondFactorType(), $this->skipProvePossessionSecondFactorTypes)) {
+            return true;
+        }
+
+        return false;
```
MKodde, line 58: Maybe turn this into a one liner?

src/Surfnet/Stepup/Identity/Entity/VerifiedSecondFactor.php
```diff
@@ -138,8 +139,26 @@ public function canBeVettedNow()
         );
     }
 
-    public function vet(DocumentNumber $documentNumber)
+    public function vet(DocumentNumber $documentNumber, $provePossessionSkipped)
     {
+        if ($provePossessionSkipped) {
+            $this->apply(
+                new SecondFactorVettedPossessionSkippedEvent(
```
MKodde, line 146: I see the logic in not adding a flag to the existing SecondFactorVettedEvent as this is a clearly different variety of a regularly vetted second factor. However I find the name of this event rather indistinct. I do not directly have a more fitting proposal, but this might be something we want to discuss with our PO as this kind of naming should match our ubiquitous language.

src/Surfnet/Stepup/Identity/Event/SecondFactorVettedPossessionSkippedEvent.php
```diff
@@ -0,0 +1,177 @@
+<?php
+
+/**
+ * Copyright 2014 SURFnet bv
```
MKodde, line 4: IIRC the current convention is to use the current year in the licence. 

src/Surfnet/Stepup/Helper/SecondFactorProvePossessionHelper.php also violates this.

src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/AuditLogRepository.php
```diff
@@ -47,6 +47,8 @@ public function __construct(ManagerRegistry $registry)
         'Surfnet\Stepup\Identity\Event\PhonePossessionProvenAndVerifiedEvent',
         'Surfnet\Stepup\Identity\Event\EmailVerifiedEvent',
         'Surfnet\Stepup\Identity\Event\SecondFactorVettedEvent',
+        'Surfnet\Stepup\Identity\Event\SecondFactorVettedPossessionSkippedEvent',
+        'Surfnet\Stepup\Identity\Event\SecondFactorVettedPossessionSkippedEvent',
```
MKodde, line 51: Duplicate

src/Surfnet/StepupMiddleware/ApiBundle/Controller/VerifiedSecondFactorController.php
```diff
@@ -104,4 +117,21 @@ public function collectionOfIdentityAction(Request $request)
 
         return JsonCollectionResponse::fromPaginator($paginator);
     }
+
+    public function getSkipVettingAction($id)
```
MKodde, line 121: Not sold on the name of this action. As this will return a boolean (json encoded) result, a method name starting with `is` seems more intuitive.

Some quick proposal

`isProvePossesionRequiredAction`
`isProvePossesionRequiredForAction`

pablothedude: Changed to `getCanSkipProvePossession`

src/Surfnet/Stepup/Identity/Entity/VerifiedSecondFactor.php
pablothedude: We could for example use:
`SecondFactorVettedAndTokenPossessionSkipped`
`SecondFactorVettedWithoutTokenProveOfPossession`
But currently I'm ok with using the initial name...

MKodde: I like `SecondFactorVettedWithoutTokenProveOfPossession`

However mind that this contains a typo, should be:

`SecondFactorVettedWithoutTokenProofOfPossession`

pablothedude: Agree, will change that

PR #314, opened 2020-09-21
Fix files not added to root of archive

phavekes:
I'm not shure why '*' was excluded, but it caused the new component_info was missing in the archive, and deploy failing.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -87,7 +87,6 @@
     },
     "archive": {
         "exclude": [
-            "*",
             ".*",
             "*/Tests/*",
             "!/bin",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -87,7 +87,6 @@
     },
     "archive": {
         "exclude": [
-            "*",
```
pablothedude, line 90: The trick is to first exclude all files and only add the files needed. That way we get a clean tarball with only the required files and no testing debugging or other files on production we don't want. That way the tarball is as lightweight as possible. Probably the best thing to do is  to add `!/component_info` and revert this change.

pablothedude (changes requested):

composer.json
phavekes: Thank you for your suggestion. I've made the suggested updates.

PR #320, opened 2021-03-08
Add self vet feature to Middleware

MKodde:
This option was added throughout the application. The setting will be saved to the event store, but is also projected to the institution configuration projection. This way the setting can be retrieved via the Middelware management API. At this location the option is also evaluated
to store it for both new and existing institutions. For institutions without the configuration option, the default 'false' is configured.

I later also added the actual SelfVet command, used to indicate a self vet option was used. This resulted in quite a lot of additional code changes. Maybe better to review them on a per-commit basis to keep things somewhat manageable.

Be advised! a migration was added to project the new setting. Be sure to run your Doctrine migrations command

This PR will not pass on travis on account of the security tests failing. They are patched up in #321. I do not see any security advisories on that branch. So this should be green too.

See: https://www.pivotaltracker.com/story/show/176714989
See: https://www.pivotaltracker.com/story/show/176715329


src/Surfnet/Stepup/Identity/Identity.php
```diff
@@ -495,6 +498,51 @@ public function vetSecondFactor(
...
+        $authoringSecondFactor = $this->getVettedSecondFactor($authoringSecondFactorId);
+        if ($authoringSecondFactor === null) {
+            throw new DomainException(
+                sprintf('Authorizing second factor with ID %s does not exist', $authoringSecondFactorId)
+            );
+        }
+        if ($authoringSecondFactor->hasEqualOrHigherLoaComparedTo($registeringSecondFactor, $secondFactorTypeService)) {
+            throw new DomainException("The second factor to be vetted has a higher LoA then the Token used for proving posession");
+        }
+        $registeringSecondFactor->vet(DocumentNumber::unknown(), true);
```
pablothedude, line 543: Could we refactor the use of DocumentNumber so we could use it for RV and SV to show the type of vetting?

pablothedude:
The build seems to be failing.

src/Surfnet/Stepup/Identity/Identity.php
MKodde: Yes, ended up adding a vetting type interface which can be implemented for your preferred vetting type. I added the on premise and self vet varieties. The first has the document number, and the self vet option tracks the token loa the vet action was self vetted with. You should be able to create a similar remote vet option

MKodde:
> When do you want to add a endpoint which could be checked if SV is possible like the optional prove possession step to keep MW as SSOT?

At this point the MW institution config is acting as the decision if the self-vetting action is allowed, later MW is consulted on the amount of vetted second factors. This data is then combined in self service to show the self-vet option or not. This could be combined in a single action on the api controller. But I see little benefit in this.

src/Surfnet/Stepup/Identity/Identity.php
```diff
@@ -495,6 +502,53 @@ public function vetSecondFactor(
         );
     }
 
+    /**
+     * Self vetting, is when the user uses it's own token to vet another.
+     *
+     * Here the new token should have a lower or equal LoA to that of the one in posession of the user
+     */
+    public function selfVetSecondFactor(
+        Loa $authoringSecondFactorLoa,
```
pablothedude, line 511: Could you elaborate why the id of the authoring second factor isn't used to query the Collection in $this->vettedSecondFactors for the required LOA?

src/Surfnet/Stepup/Identity/Value/OnPremiseVettingType.php
```diff
@@ -0,0 +1,57 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Surfnet\Stepup\Identity\Value;
+
+class OnPremiseVettingType extends VettingType
```
pablothedude, line 21: Why not make the VettingType an interface to get rid of the inheritance?

src/Surfnet/Stepup/Identity/Value/VettingType.php
```diff
@@ -0,0 +1,50 @@
...
+
+
+    abstract public function auditLog(): string;
+
+    public function jsonSerialize(): array
+    {
+        return ['type' => $this->type()];
+    }
+
+    public function type(): string
```
pablothedude, line 41: Please make this method abstract to enforce the implementation in the children.

src/Surfnet/Stepup/Identity/Value/VettingType.php
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * @var string
+     */
+    protected $type;
+
+
+    abstract public function auditLog(): string;
+
+    public function jsonSerialize(): array
```
pablothedude, line 36: Please make this method abstract to enforce the implementation in the children.

src/Surfnet/StepupMiddleware/ApiBundle/Identity/Projector/AuditLogProjector.php
```diff
@@ -91,6 +99,11 @@ private function applyAuditableEvent(AuditableEvent $event, DomainMessage $domai
             $entry->actorId         = $metadata['actorId'];
             $entry->actorCommonName = $actor->commonName;
         }
+        if ($event instanceof SecondFactorVettedEvent ||
```
pablothedude, line 102: Why isn't this coming from the metadata of the event? It seems to me that the responsibility of the metadata is to do this.

src/Surfnet/StepupMiddleware/ApiBundle/Identity/Projector/RaSecondFactorProjector.php
```diff
@@ -262,7 +263,12 @@ public function applySecondFactorVettedEvent(SecondFactorVettedEvent $event)
     {
         $secondFactor = $this->raSecondFactorRepository->find((string) $event->secondFactorId);
 
-        $secondFactor->documentNumber = $event->documentNumber;
+        $documentNumber = null;
```
pablothedude, line 266: Please move this logic to the concrete implementations of the VettingTypes and create a static `documentNumber` method that way the concrete implementation is responsible for what it will return and you and you can get rid of this logic.

src/Surfnet/StepupMiddleware/CommandHandlingBundle/Identity/CommandHandler/IdentityCommandHandler.php
```diff
@@ -347,6 +340,33 @@ public function handleVetSecondFactorCommand(VetSecondFactorCommand $command)
...
 
+    public function handleSelfVetSecondFactorCommand(SelfVetSecondFactorCommand $command)
+    {
+        /** @var IdentityApi $identity */
+        $identity = $this->eventSourcedRepository->load(new IdentityId($command->identityId));
+        $secondFactorIdentifier = SecondFactorIdentifierFactory::forType(
+            new SecondFactorType($command->secondFactorType),
+            $command->secondFactorId
+        );
+        $loa = $this->loaResolutionService->getLoa($command->authoringSecondFactorIdentifier);
```
pablothedude, line 351: Domain logic should reside in aggregates and not in command handlers, please move this to the Identity aggregate.

src/Surfnet/StepupMiddleware/CommandHandlingBundle/Tests/SensitiveData/SensitiveDataTest.php
```diff
@@ -61,17 +61,6 @@ public function sensitiveDataToSerialise()
                     ->forget(),
                 ['CommonName' => CommonName::unknown(), 'Email' => Email::unknown()],
             ],
-            'DocumentNumber' => [
```
pablothedude, line 64: Shouldn't this be tested with the `VettingType` object / `withVettingType` method?
And are you sure the sensitive data is going to work with the old event stream? Please test this with the old `document_number` set as sensitive data.

src/Surfnet/StepupMiddleware/CommandHandlingBundle/SensitiveData/SensitiveData.php
```diff
@@ -180,12 +178,13 @@ public static function deserialize(array $data)
...
     public function serialize(): array
     {
+        $vettingType = (!is_null($this->vettingType)) ? $this->vettingType->jsonSerialize() : null;
         return array_filter([
             'common_name'              => $this->commonName,
             'email'                    => $this->email,
             'second_factor_type'       => $this->secondFactorType,
             'second_factor_identifier' => $this->secondFactorIdentifier,
-            'document_number'          => $this->documentNumber,
+            'vetting_type' => $vettingType
```
pablothedude, line 187: Shouldn't `document_number` not be supported, this seems a BC breaking change to me?

src/Surfnet/StepupMiddleware/ApiBundle/Tests/Doctrine/Type/SelfVetOptionTypeTest.php
```diff
@@ -0,0 +1,122 @@
+<?php
+
+/**
+ * Copyright 2018 SURFnet B.V.
```
phavekes, line 4: +3

public/index.php
```diff
@@ -12,6 +12,11 @@
     Debug::enable();
 }
 
+if (isset($_COOKIE['testcookie']) && strpos($_SERVER['HTTP_USER_AGENT'], 'GuzzleHttp') !== false) {
```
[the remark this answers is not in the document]
MKodde: This can be removed from the `index.php` as per: `https://github.com/OpenConext/Stepup-Deploy/pull/135` Your concern seems to be covered in that PR. As only outside of the prod mode this check will be honored

src/Surfnet/Stepup/Identity/Event/SecondFactorVettedEvent.php
```diff
@@ -124,25 +115,43 @@ public function getAuditLogMetadata()
...
+        $metadata->vettingType = $this->vettingType;
 
         return $metadata;
     }
 
     public static function deserialize(array $data)
     {
         $secondFactorType = new SecondFactorType($data['second_factor_type']);
-
+        $vettingType = new UnknownVettingType();
```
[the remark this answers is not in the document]
MKodde: I see your point, but leaving this as is for now

src/Surfnet/Stepup/Identity/Identity.php
MKodde: Sadly we have no knowledge of the authoring second factor other then the LoA it had. Gateway does not store the token ID used in it's assertion back to the SP (SelfService).

src/Surfnet/Stepup/Identity/Value/OnPremiseVettingType.php
MKodde: Mainly out of convenience: the type and __toString methods would always be the same.

src/Surfnet/StepupMiddleware/CommandHandlingBundle/Identity/CommandHandler/IdentityCommandHandler.php
MKodde: I misused the domain exception here. I don't view this as domain logic, but more like input validation/input dressing. The alternative would be to inject the loa resolution helper onto the identity aggregate, and that would introduce other undesirable issues.

src/Surfnet/StepupMiddleware/ApiBundle/Identity/Projector/AuditLogProjector.php
MKodde: Please elaborate

src/Surfnet/Stepup/Identity/Value/VettingType.php
MKodde: Resolved this by moving to a composition based solution.

src/Surfnet/StepupMiddleware/ApiBundle/Identity/Projector/AuditLogProjector.php
pablothedude: All loggable information currently comes from the metadata of the event. So by just checking if the metadata exists you prevent coupling with concrete Event implementations in the AuditlogProjector. You could also get rid of the VettingType::auditLog method because this auditlog logic doesn't seem to belong in this VO.

phavekes:
Something's wrong with the email-template:
`Apr 26 11:14:50 app stepup-selfservice[1304]: {"channel":"app","level":"WARNING","message":"Command 'Identity:ProveGssfPossession' with UUID 'd9f43c86-19c4-4278-acb0-34694df16bb8' could not be executed (Twig\\Error\\LoaderError: Template reference \"SurfnetStepupMiddlewareCommandHandlingBundle:SecondFactorMailService:email.html.twig\" not found, did you mean \"@SurfnetStepupMiddlewareCommandHandling/SecondFactorMailService/email.html.twig\"?)","context":[],"extra":{"server":"selfservice.stepup.example.com","application":"self-service","request_id":"64119cd2b8e0267f6226af426f4c700e"}}`

src/Surfnet/StepupMiddleware/ApiBundle/Identity/Projector/AuditLogProjector.php
MKodde: Good point, I completely ovetlooked this. The AuditableEvent can be used neatly for this intent. For now I will leave the retrieval of the audit log entry on the VO. I feel it is the responsibility of the vetting type to determine how its audit log entry is composed. Making this the responsibility of the projector is also not a pure responsibility of the projector in my opinion.

src/Surfnet/StepupMiddleware/CommandHandlingBundle/SensitiveData/SensitiveData.php
MKodde: Document number is implicitly supported in the vetting type. I believe this is tried and tested. But maybe I'm missing a test scenario here.

My tests included: event replaying with old/new style events. And performed the events from the applications.

MKodde:
@pablothedude I addressed most of your feedback. Some points where not picked up as I found the retrurn on investment not worth it. I found it hard to run ample tests on the sensitive data do you have some tips or tricks to dive a bit deeper into them?

MKodde:
> Something's wrong with the email-template:
> `Apr 26 11:14:50 app stepup-selfservice[1304]: {"channel":"app","level":"WARNING","message":"Command`

Hi @phavekes, ea8fed7 should fix your issues. Strangely I did not run across this issue. Maybe my environment does not run in strict mode or another similar glitch was encountered. Can you verify this fixes the issue at hand?

phavekes:
> > Something's wrong with the email-template:
> > `Apr 26 11:14:50 app stepup-selfservice[1304]: {"channel":"app","level":"WARNING","message":"Command`
> 
> Hi @phavekes, [ea8fed7](https://github.com/OpenConext/Stepup-Middleware/commit/ea8fed7747460d0fc3e17d6cb3de7abba7c6807a) should fix your issues. Strangely I did not run across this issue. Maybe my environment does not run in strict mode or another similar glitch was encountered. Can you verify this fixes the issue at hand?

This issue is fixed

src/Surfnet/StepupMiddleware/CommandHandlingBundle/Tests/SensitiveData/SensitiveDataTest.php
pablothedude: Did you validate this?

src/Surfnet/StepupMiddleware/CommandHandlingBundle/SensitiveData/SensitiveData.php
pablothedude: Seems OK then.

pablothedude:
> @pablothedude I addressed most of your feedback. Some points where not picked up as I found the return on investment not worth it. I found it hard to run ample tests on the sensitive data do you have some tips or tricks to dive a bit deeper into them?

I think the best is to test this by forgetting an identity then then check if all sensitive data is forgotten too.

src/Surfnet/StepupMiddleware/CommandHandlingBundle/Tests/SensitiveData/SensitiveDataTest.php
MKodde: Yup, tested that, and all seems to work as intended.

MKodde:
At this point we have not yet got a construction to forget about an identity. The even replay however seems to work correctly on the old and new style events. So I'm quite certain all works like it should!

src/Surfnet/StepupMiddleware/ApiBundle/Identity/Projector/AuditLogProjector.php
```diff
@@ -91,6 +100,9 @@ private function applyAuditableEvent(AuditableEvent $event, DomainMessage $domai
             $entry->actorId         = $metadata['actorId'];
             $entry->actorCommonName = $actor->commonName;
         }
+        if (property_exists($auditLogMetadata, 'vettingType') && !is_null($auditLogMetadata->vettingType)) {
```
Badlapje, line 103: Why not transfer the if to a helper function "hasVettingType"?  Would make it easier to read.

Badlapje: In fact, you could just create a helper method to add to the actorCommonname conditionally.  That should also reduce the cyclomatic complexity meaning you can get rid of the warning suppression.

Something like...

```PHP
private function enhanceActorCommonNameWithAuditLog ($entry, $auditLogMetadata) {
  if (property_exists($auditLogMetadata, 'vettingType') && !is_null($auditLogMetadata->vettingType)) {
      return $entry->actorCommonName . $auditLogMetadata->vettingType->auditLog();
  }
  return $entry->actorCommonName
}
```

src/Surfnet/StepupMiddleware/ApiBundle/Tests/Authorization/Service/CommandAuthorizationServiceTest.php
```diff
@@ -314,7 +315,10 @@ public function an_identity_should_not_be_able_to_execute_someone_elses_selfserv
                 ->andReturn($raCredentials);
 
             // These commands should always be allowed to be executed
-            if ($command instanceof CreateIdentityCommand || $command instanceof UpdateIdentityCommand) {
+            if ($command instanceof CreateIdentityCommand ||
```
Badlapje, line 318: Maybe also extract this If condition to a seperate method?  Function alwaysExecutableCommand or something like that.  That'd allow to get rid of the comment.

Badlapje (approved):
Some nitpicks, but looks good to me other than those: nice work!

src/Surfnet/StepupMiddleware/ApiBundle/Identity/Projector/AuditLogProjector.php
MKodde: I like this suggestion. Will pickitup

src/Surfnet/StepupMiddleware/ApiBundle/Tests/Authorization/Service/CommandAuthorizationServiceTest.php
MKodde: Given this is test code, I feel comfortable leaving this as is.

PR #326, opened 2021-06-01
Introduce the Move token console command

MKodde:
This includes the creation of the console command, the fixing up of the
event handling/command handling. And finally the projectors have been
updated to correctly project the changes.

https://www.pivotaltracker.com/story/show/178092669


src/Surfnet/StepupMiddleware/ApiBundle/Identity/Projector/SecondFactorProjector.php
```diff
@@ -234,7 +235,9 @@ public function applySecondFactorVettedEvent(SecondFactorVettedEvent $event)
         $vetted->secondFactorIdentifier = $event->secondFactorIdentifier->getValue();
 
         $this->vettedRepository->save($vetted);
-        $this->verifiedRepository->remove($verified);
+        if ($verified && !$event->vettingType instanceof MoveTokenVettingType) {
```
pablothedude, line 238: Is it not better to just always remove a token if it exists, no matter what the type is? The state should already be validated in the aggregate, so it's safe to assume you can always remove the token.

pablothedude (changes requested):
A nice feature, please see address my comments if you see fit.

PR #327, opened 2021-06-03
Remove vetting type from event stream

MKodde:
The details can be found here: https://www.pivotaltracker.com/story/show/178390934


MKodde:
Given this is to be backported to the current version. I did not add a fix for the failing security check. If we need this patched in the current version we can do so in another PR.

phavekes (approved):

pablothedude (approved):

PR #328, opened 2021-07-02
Project token migration audit log data

pablothedude:
https://www.pivotaltracker.com/story/show/178390934


pablothedude:
Will be fixed in https://github.com/OpenConext/Stepup-Middleware/pull/326

PR #329, opened 2021-07-02
Add sensitive data cleanup migration

pablothedude:


src/Surfnet/Migrations/Version20210610131957.php
```diff
@@ -0,0 +1,68 @@
...
+ * removes the sensitive data via JSON encoding/decoding the payload of the event.
+ */
+final class Version20210610131957 extends AbstractMigration
+{
+    private static $select = <<<SQL
+        SELECT uuid, playhead, payload
+        FROM event_stream
+        WHERE (type = 'Surfnet.Stepup.Identity.Event.SecondFactorVettedEvent'
+            OR type = 'Surfnet.Stepup.Identity.Event.SecondFactorVettedWithoutTokenProofOfPossession')
+        AND recorded_on > '2021-04-11'
```
phavekes, line 22: Should we check for the presence of document_number, instead of a recorded_on? The date is very SURF-specific, and a document number should never be present in the event_stream.

phavekes (changes requested):

src/Surfnet/Migrations/Version20210610131957.php
pablothedude: In the line below it's checking for the `vetting_type` which is the element which holds the 'document_number'. That element should be removed completely, also for other types of VettingTypes. So that's why the `vetting_type` is being used. The `recorded_on` is used to only migrate events after the date were the bug was introduced to prevent iterating unnecessary all events. In the original change a between was used on the `recorded_on` data and I did remove that in order to support other users who could potentially run the migration later in time.

src/Surfnet/Migrations/Version20210610131957.php
```diff
@@ -0,0 +1,71 @@
...
+                    $uuid,
+                    $playhead
+                )
+            );
+        }
+    }
+
+    public function down(Schema $schema) : void
+    {
+        // This migration can not be undone.
```
Badlapje, line 62: ❤️

Badlapje (approved):
Code is nice and clean, nice work!

PR #330, opened 2021-07-08
Allow vetting_type sensitive data deserialization

pablothedude:
There was a fix to prevent the vetting_type from going into the normal
event stream. However the vetting_type sensitive data was now being
deserialized back into the event because this logic wasn't yet moved
to the SensitiveData implementation in that fix.


Badlapje (approved):
Code-wise i have no remarks 💯

PR #336, opened 2021-11-22
 Improve type safety by setting sql_mode to be strict part 2

MKodde:
Merged a bit to quickly, I addressed two review points from @pmeulen 

Part one can be found here, including the feedback mentioned above here #335


pmeulen (approved):
👍

PR #339, opened 2022-02-01
Reproduce buggy behavior in useRaa configuration

MKodde:
When an institution is configured to use RAA's from another institution,
but the useRa is not set for that same institution. And a RAA user is
made RAA for that institution. He/She is not allowed to perform RA
actions (vet second factor and revocation of a token).

This is because in the authorisation service we 'downgrade' the Vet and
Revoke actions to require the use ra role.

The succesive buildInstitutionsAuthorizationContext then searches for
the institutions that have that useRa role applied. None are found, an
authorization exception is raised.

To fix this. Either add both the useRa and useRaa actions to the authz
context. Or make the parent method more user context aware. Where we
would have to check that if the useRaa role is applied for either one of
the Vet or Revoke events, we should set the useRaa role instead of the
hard coded useRa.

See: https://www.pivotaltracker.com/story/show/180081100


MKodde:
This is not yet fixed in: 

https://github.com/OpenConext/Stepup-Middleware/pull/351
https://www.pivotaltracker.com/story/show/181537313

But it did touch on a similar issue. The issue at hand here is that the RAA role is not guaranteed to allow execution of RA actions.

PR #342, opened 2022-03-15
Add a deprovision endpoint

MKodde:
Creates:
- The deprovision user
- Deprovision controller with two actions (deprovision and dry run)
- Router configuration for the aforementioned actions
- Updated API docs
- Two deprovision calls in the Postman collection

### DIFF
--- a/config/legacy/parameters.yaml.dist
+++ b/config/legacy/parameters.yaml.dist
@@ -43,10 +43,12 @@ parameters:
     # - registration - user "ra"
     # - readonly access to all endpoints - user "apireader"
     # - management - user "management"
+    # - GDPR complience: deprovision and retrieval of user information - user "lifecycle"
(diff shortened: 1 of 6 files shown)


config/legacy/parameters.yaml.dist
```diff
@@ -43,10 +43,12 @@ parameters:
     # - registration - user "ra"
     # - readonly access to all endpoints - user "apireader"
     # - management - user "management"
+    # - deprovision - user "deprovision"
```
phavekes, line 46: Will the same user be used for the userinformation endpoint? In that case, we should choose another name

MKodde: The user is now named: lifecycle, the parameter to set its password is updated to: `lifecycle_password`

MKodde (commented):
Thanks for your review @phavekes

VadimSchmitz (approved):

PR #345, opened 2022-03-22
Install Composer updates

MKodde:
Aiming to be up to date within our package version restraints, and also to remedy the security warning that was bugging us the previous PR's

Amongst the updates is Symfony Flex. It was upgraded from 1.13 to v1.18. I looked into upgrading to Flex 2, but the PHP 8 constrained stopped me from applying that change. @tvdijen this should help you out right? Judging from your comment on #337.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -8,16 +8,16 @@
     "packages": [
         {
             "name": "beberlei/assert",
-            "version": "v3.3.1",
+            "version": "v3.3.2",
             "source": {
                 "type": "git",
                 "url": "https://github.com/beberlei/assert.git",
-                "reference": "5e721d7e937ca3ba2cdec1e1adf195f9e5188372"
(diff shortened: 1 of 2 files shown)


MKodde:
> Uhm, I'm not sure what `composer/pcre` and `symfony/polyfill-php81` have to do with the warnings that bugged me

They do not relate to your problem indeed. They are just a by product of running a composer update action. Some of the symfony recepies got updated and the symfony lockfile was updated accordingly.


> **Update:** So, to loose the composer issues, I added the `allow-plugins` and to loose the deprecation warnings (and the fact that I can't access the flex-endpoint through our proxy) I ended up setting `export FLEX_SERVERLESS=true`

I believe setting the FLEX_SERVERLESS env-var is no longer a requirement as of this version of Flex (as it became the default after 1.16 (from top of my head).

PR #348, opened 2022-03-24
Integrate event right to obtain data

MKodde:
Some major changes where applied:

1. The recent changes to the events (allowlist on data retrievable during user-data retrival) have been integrated
2. The allowlist filtering logic was removed from the events
3. A dedicated filter was created that is responsible for filtering
4. A formatter was added to create JSON serializable data from the event data

See https://www.pivotaltracker.com/story/show/181430595


src/Surfnet/StepupMiddleware/CommandHandlingBundle/SensitiveData/RightToObtainDataInterface.php
```diff
@@ -4,5 +4,17 @@
 
 interface RightToObtainDataInterface
 {
+    /**
```
pablothedude, line 7: :100:

src/Surfnet/StepupMiddleware/CommandHandlingBundle/SensitiveData/RightToObtainDataInterface.php
```diff
@@ -4,5 +4,17 @@
...
+     * Obtains the user data, suitable for showing a user that exercised its
+     * right to obtain user data.
+     */
     public function obtainUserData(): array;
+
+    /**
+     * Retrieve the list of allowed data to retireve from the event.
```
pablothedude, line 14: :100:

src/Surfnet/Stepup/Helper/UserDataFormatter.php
```diff
@@ -0,0 +1,59 @@
...
+class UserDataFormatter implements UserDataFormatterInterface
+{
+    private $applicationName;
+
+    public function __construct(string $applicationName)
+    {
+        $this->applicationName = $applicationName;
+    }
+
+    public function format(array $userData, array $errors): array
```
pablothedude, line 29: You could return a simple response DTO to detect breaking changes early on when changing code without introducing side effects and being unaware of this.

src/Surfnet/Stepup/Identity/Event/SecondFactorMigratedToEvent.php
```diff
@@ -146,8 +146,11 @@ public function obtainUserData(): array
     {
         $serializedPublicUserData = $this->serialize();
         $serializedSensitiveUserData = $this->getSensitiveData()->serialize();
```
pablothedude, line 148: You could also move this to the DataFilter and check if the Event implements the Forgettable interface and then merge the sensitive data, but feel free to leave it as-is.

pablothedude (approved):

src/Surfnet/Stepup/Identity/Event/SecondFactorMigratedToEvent.php
MKodde: That might be even better in retrospect. Making the interface more of a marker interface (save for the getter for the allowList). I will consider changing this.

src/Surfnet/Stepup/Helper/UserDataFormatter.php
MKodde: That was one of the things @VadimSchmitz and I discussed in our whiteboard session. In the end I felt the added benefit you describe was not quite worth the added 'complexity'. I will reconsider our design decisions, and perhaps update the code accordingly.

VadimSchmitz (approved):
Good work on implementing what we discussed.

PR #349, opened 2022-03-28
Convert TravisCi tests to a GithubActions workflow

VadimSchmitz:
This commit fully converts Stepup-Middleware from TravisCi  & Ant scripts to GithubActions and Composer scripts.

Work done to make this conversion:
- Move each Ant task into its own scripts.
- Docker test container is used to run the tests in.
- GithubActions workflow is configured.
- All TravisCi and Ant dependencies are removed.

Currently PHPCPD has been disabled due to Xdebug throwing a fatal error. this could be fixed in the future by updating PHP so PHPCPD can be updated too, In the meantime Scrutinizers tools can be used to do the same thing. 

[181446825](https://www.pivotaltracker.com/story/show/181446825)


.gitignore
```diff
@@ -30,7 +30,7 @@ config/services.yml
 
 ###> squizlabs/php_codesniffer ###
 /.phpcs-cache
-/phpcs.xml
+/ci/qa/phpcs.xml
```
MKodde, line 33: Why do you want to git-ignore the phpcs config file? Don't be fooled by the boilerplate code squizlabs put in our `.gitignore`

.github/workflows/test-integration.yml
```diff
@@ -0,0 +1,34 @@
...
+      - name: Install dependencies
+        run: |
+          cd ci/docker && docker-compose exec -T stepup-middleware bash -lc '
+          composer install
+          '
+      - name: Run test scripts
+        run: cd ci/docker && docker-compose exec -T stepup-middleware bash -lc ' composer test '
+      - name: Output log files on failure
+        if: failure()
+        run: cd ci/docker && docker-compose exec -T stepup-middleware -c 'cat var/log/webtest.log'
```
MKodde, line 34: This will not work here. You are explicitly running tests in the `test` env. Please update, and test if the log file shows useful information when an erroneous situation occurs.

composer.json
```diff
@@ -43,20 +43,22 @@
         "symfony/mailer": "4.4.*",
         "symfony/maker-bundle": "^1.19",
         "symfony/monolog-bundle": "^3.5",
-        "symfony/profiler-pack": "^1.0",
```
MKodde, line 46: I wonder why the profiler-pack was removed here? Was it replaced by the web-profiler-bundle?

composer.json
```diff
@@ -43,20 +43,22 @@
...
         "symfony/security-bundle": "4.4.*",
         "symfony/swiftmailer-bundle": "^3.4",
         "symfony/translation": "4.4.*",
-        "symfony/twig-pack": "^1.0",
```
MKodde, line 50: Again, why was this updated? Can this be a composer 2 doing?

composer.json
```diff
@@ -75,6 +79,21 @@
...
+            "@phpcpd",
+            "@phpunit",
+            "@security-tests"
+        ],
+        "phplint": "./ci/qa/phplint",
+        "phpmd": "./ci/qa/phpmd",
+        "phpcs": "./ci/qa/phpcs",
+        "phpcpd": "./ci/qa/phpcpd",
+        "phpunit": "./ci/qa/phpunit",
+        "security-tests": "./ci/qa/security-tests",
```
MKodde, line 95: :100:

composer.lock
```diff
@@ -11393,5 +11486,5 @@
     "platform-overrides": {
         "php": "7.2"
     },
-    "plugin-api-version": "1.1.0"
+    "plugin-api-version": "2.0.0"
```
MKodde, line 11489: Gotcha! You used composer 2 to install the dependencies. Please revert the changes to the composer.json that are not in place, and re-install using composer 1. And finally commit that lockfile in your PR.

ci/qa/phplint
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+
+cd $(dirname $0)/../../
+mkdir -p var/qa
+
+# https://github.com/overtrue/phplint
+./vendor/bin/phplint --configuration=ci/qa/phplint.yaml $1
```
MKodde, line 8: Out of curiosity, why did you not re-purpose the previous PHP lint? 

Was it hard to get a list of changed files to get this to work?
`sh -c "echo '${changeset.php.spacesep}' | xargs -n 1 -P 4 php -l 1>/dev/null"`

I have little objection to using ta linter tool that can do this work for us. Have you looked at other linting machines that can also lint our other stuff?

Also have a look in the SF console: 

```
 lint
  lint:container                              Ensure that arguments injected into services match type declarations
  lint:twig                                   Lint a template and outputs encountered errors
  lint:xliff                                  Lint an XLIFF file and outputs encountered errors
  lint:yaml                                   Lint a file and outputs encountered errors
```

MKodde (changes requested):
Nice work! 

I found some code review points and have the following questions 

Can these be removed? They seem to be unused, please verify:
- ci/qa/phpcs.xml.dist
- ci/qa/phpmd-pre-commit.xml
- config/packages/security_checker.yaml

ci/qa/phpcpd
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+#cd $(dirname $0)/../../
+#
+## https://github.com/sebastianbergmann/phpcpd
+#./vendor/bin/phpcpd --exclude src/*/Tests/* src
+
+# PHPCS is currently disabled due to Xdebug throwing a fatal error.
+echo "Skipping PHPCPD due to Xdebug fatal error"
```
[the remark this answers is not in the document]
VadimSchmitz: There are no JS dependencies in this repository. so it would mean that a bunch of yarn boilerplate code would be added for just one library. Scrutinize should do the same as PHPCPD.

composer.json
VadimSchmitz: This is the same reason as previously mentioned.
see: https://packagist.org/packages/symfony/twig-pack

composer.lock
VadimSchmitz: It seems like this was the only change within the composer.lock file.
The .lock file has been regenerated in the new commit.

ci/qa/phplint.yaml
```diff
@@ -0,0 +1,7 @@
+path: [./src, ./tests]
+jobs: 10
+cache: /var/qa/phplint.cache
```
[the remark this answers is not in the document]
VadimSchmitz: /var/ is already part of the .gitignore file.

ci/qa/phplint
VadimSchmitz: I personally felt that the previous lint was less readable arguments wise and the other projects also use this overtrue/phplint.
If needed  the old one can be implemented again.

.github/workflows/test-integration.yml
VadimSchmitz: has been changed to the correct var/log/dev.log file

composer.json
VadimSchmitz: Due to a composer require all the other packages are updated too. The update will bring no other changes next to the symfony/profiler-pack getting deleted. This is because the other packages are part of the pack and are also installed.

composer.json
VadimSchmitz: Due to a composer require all the other packages are updated too. The update will bring no other changes next to the symfony/twig-pack getting deleted. This is because the other packages are part of the pack and are also installed.

.github/workflows/test-integration.yml
```diff
@@ -0,0 +1,34 @@
...
+      - name: Install dependencies
+        run: |
+          cd ci/docker && docker-compose exec -T stepup-middleware bash -lc '
+          composer install
+          '
+      - name: Run test scripts
+        run: cd ci/docker && docker-compose exec -T stepup-middleware bash -lc ' composer test '
+      - name: Output log files on failure
+        if: failure()
+        run: cd ci/docker && docker-compose exec -T stepup-middleware -c 'cat var/log/dev.log'
```
MKodde, line 34: It's highly unlikely this test is running in the `dev` env. I wanted to verify this myself yesterday, but ran out of time. Can you check this? And maybe link a GHA run where the build failed and the test log is successfully catted.

composer.json
MKodde: Yesterday I learned: Composer require == composer update + composer require.. Thanks for looking into this!

MKodde (changes requested):
Two issues remain. Almost there!

.github/workflows/test-integration.yml
VadimSchmitz: As discussed the logging will be parked for now.

ci/qa/phpmd.xml
```diff
@@ -32,6 +32,7 @@
         <exclude name="LongVariable" />
         <exclude name="LongClassName" />
         <exclude name="ShortClassName" />
+        <exclude name="ShortMethodName" />
```
MKodde, line 35: This one can be removed IMO, unless you have a valid reason for adding it of course!

MKodde (changes requested):
Almost there! Please fix the last remaining issue!

ci/qa/phpcpd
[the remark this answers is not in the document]
MKodde: I can agree with your reasoning. However we have Yarn on the test container. But I agree, having a CI only test is not very predictable. And having one single dev-only JS dependency is not that cute either..

ci/qa/phpmd.xml
VadimSchmitz: It was not in the previous test either, therefore it can be removed

ci/qa/phpmd
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+cd $(dirname $0)/../../
+
+./vendor/bin/phpmd src text ci/qa/phpmd.xml --exclude */Tests/*,*Migrations/*
```
MKodde, line 5: :+1:

MKodde (approved):

PR #351, opened 2022-04-05
Repair broken authz in mayRaCommandBeExecutedOnBehalfOf

MKodde:
When vetting a token, the authorization check was not sharp enough. (incomplete)
1. The actor user's institution was only used to test for which other institutions it is allowed to perform tasks for. This would work for any configuration where 'regular' use-ra(a) configuration where set for that institution. When the RA is selected for a different institution, that test would not get the authorizations for that institution but just grab those of the 'home institution' of the actor. Adding a second query to also include the 'select_raa' authorizations for that institution fixed this problem.
2. To debug and get a grasp of what is happening in the Autz service. Quite some logging was added. The Autz repo also was exteded with some additional logging.

See https://www.pivotaltracker.com/story/show/181537313

### DIFF
--- a/src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/AuthorizationRepository.php
+++ b/src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/AuthorizationRepository.php
@@ -21,6 +21,7 @@ namespace Surfnet\StepupMiddleware\ApiBundle\Identity\Repository;
 use Doctrine\Bundle\DoctrineBundle\Repository\ServiceEntityRepository;
 use Doctrine\Common\Persistence\ManagerRegistry;
 use Doctrine\ORM\Query\Expr\Join;
+use Psr\Log\LoggerInterface;
 use Surfnet\Stepup\Configuration\Value\InstitutionRole;
 use Surfnet\Stepup\Identity\Collection\InstitutionCollection;
 use Surfnet\Stepup\Identity\Value\IdentityId;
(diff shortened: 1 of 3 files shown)


src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/AuthorizationRepository.php
```diff
@@ -73,12 +84,38 @@ public function getInstitutionsForRole(InstitutionRole $role, IdentityId $actorI
...
+            $this->logger->notice(
+                sprintf('Adding %s to authorized institutions', $institution['institution'])
+            );
             $result->add(new Institution((string)$institution['institution']));
         }
 
+        // Also get the institutions that are linked to the user via the 'institution_relation' field.
+        // Effectively getting the use_raa relation.
+        // See https://www.pivotaltracker.com/story/show/181537313
+        $qb = $this->_em->createQueryBuilder()
```
pablothedude, line 97: Would it be sufficient to fix the where clause in the query above?

src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/AuthorizationRepository.php
```diff
@@ -32,11 +33,20 @@
...
-    public function __construct(ManagerRegistry $registry)
+    /**
+     * @var LoggerInterface
+     */
+    private $logger;
+
+    public function __construct(ManagerRegistry $registry, LoggerInterface $logger)
     {
         parent::__construct($registry, AuditLogEntry::class);
+        $this->logger = $logger;
```
pablothedude, line 49: Logging :100:

pablothedude (approved):

src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/AuthorizationRepository.php
MKodde: I have tried this after your suggestion. And added an additional join condition to the existing query. That did not result in the correct outcome.

For some extra context, query 1:
Selects the authorized institutions that apply for the home institution of the RA that is interacting with Middleware (the ra_institution)

And query 2:
Checks the Institution the RA has been selected for (using the SELECT_RA feature). And selects in which institutions the RA has RA(A) rights.

MKodde:
Thanks for your review @pablothedude. 

Shipping :ship:

src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/AuthorizationRepository.php
pablothedude: :+1: thanks for validating!

PR #352, opened 2022-05-04
Replace deprecated Swiftmailer with Symfony mailer

VadimSchmitz:
All mail services that used the deprecated Swiftmailer are not using Symfony mailer.

https://www.pivotaltracker.com/story/show/179976844

### DIFF
--- a/.github/workflows/test-integration.yml
+++ b/.github/workflows/test-integration.yml
@@ -2,7 +2,7 @@ name: test-integration
 on:
   pull_request:
   push:
-    branches: [ master, develop ]
+    branches: [ master, develop, feature/*, bugfix/* ]
 jobs:
   build:
     runs-on: ubuntu-latest
--- a/composer.lock
+++ b/composer.lock
@@ -5378,16 +5378,16 @@
         },
         {
             "name": "symfony/mailer",
-            "version": "v4.4.38",
+            "version": "v4.4.41",
             "source": {
                 "type": "git",
                 "url": "https://github.com/symfony/mailer.git",
-                "reference": "cb843f81cbe16bfd2e17e23d619adb2dddfd0bde"
+                "reference": "62d909879078b31f338b26a2701c8f8802343769"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/symfony/mailer/zipball/cb843f81cbe16bfd2e17e23d619adb2dddfd0bde",
-                "reference": "cb843f81cbe16bfd2e17e23d619adb2dddfd0bde",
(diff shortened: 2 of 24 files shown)


VadimSchmitz:
Since i cannot test the Symfony mailer Michiel Kodde will finish this task.
Work that needs to be done:
- Finish setting up the configuration for the global, test, smoketest, smoketest_event_replay and prod_event_replay environments.
- Another issue that arose is a coupling warning from phpmd

composer.lock
```diff
@@ -5378,16 +5378,16 @@
         },
         {
             "name": "symfony/mailer",
-            "version": "v4.4.38",
+            "version": "v4.4.41",
```
MKodde, line 5381: Cool to see the symfony mailer was already in the project. 
Did you intentionally not add it to the composer.json file?

src/Surfnet/StepupMiddleware/CommandHandlingBundle/Identity/Service/EmailVerificationMailService.php
```diff
@@ -84,7 +80,6 @@ public function __construct(
         Mailer $mailer,
         Sender $sender,
         TranslatorInterface $translator,
-        Environment $twig,
```
MKodde, line 87: You will need to update the service definitions according to this change (also goes for the other services in this commit)
The Twig environment needs to be removed from the definitions you'll find for these services..

MKodde (changes requested):
Nice work, great to see you making very logical atomic commits on this one :+1: 

Please address this feedback below, after which I will do a functional test.

composer.lock
VadimSchmitz: It is currently also in the composer.json file.

config/packages/dev/swiftmailer.yaml
```diff
@@ -1,5 +0,0 @@
-# See https://symfony.com/doc/current/email/dev_environment.html
-swiftmailer:
-    # send all emails to a specific address
-    #delivery_addresses: ['me@example.com']
-    port: 1025
```
MKodde, line 5: You removed the three swiftmailer configs, I'd expect new (framework) config settings for these environments that reflect the settings that where configured before.

.github/workflows/test-integration.yml
```diff
@@ -2,7 +2,7 @@ name: test-integration
 on:
   pull_request:
   push:
-    branches: [ master, develop ]
+    branches: [ master, develop, feature/self-asserted-tokens ]
```
tvdijen, line 5: You know that instead of keeping a list, you can use wildcards here? i.e. `feature/*`

MKodde: Good suggestion!

MKodde (approved):

PR #358, opened 2022-06-08
Implement revocation of Recovery Tokens

MKodde:
This commit implements the two varieties of token revokation, one is triggered by the Identity from the self service. The other is a RA invoked event.

See task 5 of https://www.pivotaltracker.com/story/show/181934073


src/Surfnet/StepupMiddleware/CommandHandlingBundle/Identity/Command/RevokeRegistrantsRecoveryTokenCommand.php
```diff
@@ -0,0 +1,67 @@
...
+     * @var string
+     */
+    public $recoveryTokenId;
+
+    /**
+     * @inheritDoc
+     */
+    public function getRaInstitution()
+    {
+        return null;
```
pablothedude, line 65: Does this have to be set in order to allow RA's of other organisations to do this or will this be handled in the CommandAuthorizationService?

pablothedude (approved):

PR #361, opened 2022-06-14
Create read endpoints for Recovery Tokens

MKodde:
A singular retrieve and search endpoint was added to the API. A new
controller was created that bundles the Recovery Token logic.

See Task 7 of https://www.pivotaltracker.com/story/show/181934073

### DIFF
--- a/src/Surfnet/StepupMiddleware/ApiBundle/Controller/RecoveryTokenController.php
+++ b/src/Surfnet/StepupMiddleware/ApiBundle/Controller/RecoveryTokenController.php
@@ -1,0 +1,71 @@
+<?php
+
+/**
+ * Copyright 2022 SURFnet bv
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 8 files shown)


src/Surfnet/StepupMiddleware/ApiBundle/Controller/RecoveryTokenController.php
```diff
@@ -0,0 +1,71 @@
...
+    private $service;
+
+    public function __construct(RecoveryTokenService $recoveryTokenServiceService)
+    {
+        $this->service = $recoveryTokenServiceService;
+    }
+
+    public function getAction($id)
+    {
+        $this->denyAccessUnlessGranted(['ROLE_RA', 'ROLE_SS', 'ROLE_READ']);
```
pablothedude, line 49: Could you elaborate were ROLE_READ is for?

MKodde: As the `ROLE_READ` name implies is a role that is used to access the read side of the API. IRC this is used in one of the Java apps (maybe manage or idp dashboard) to show some details about the Stepup configuration of an insittution.

PR #367, opened 2022-07-11
Event replay improvements

MKodde:
In order to play the event/projection specific event replays. Where you choose which events are replayed onto which projectors was broken and not maintained the last couple of years. 

This was addressed on this PR. The goal is to allow for an event replay on the identity self asserted token options projection. This should be slightly more lightweight than a full event replay.

~The screen cast below shows what feature I'm talking about.~ The screencast is considered to be a corrupt  mp4 file according to Github. :man_shrugging:


MKodde:
https://user-images.githubusercontent.com/28252948/178209849-aa7a32fc-c799-4df6-9e72-2ef88980ed1b.mp4

pablothedude (approved):
:crossed_fingers:

PR #368, opened 2022-07-11
Vetting type on vetted second factor projections

MKodde:
The vetting type is set on both the Gateway and Middleware Second Factor projections (for vetted second factors)

The vetting type is a relevant projection as it is used to determine the registration level of assurance (loa) of a second factor token.

### DIFF
--- a/src/Surfnet/Migrations/Version20220704084758.php
+++ b/src/Surfnet/Migrations/Version20220704084758.php
@@ -1,0 +1,28 @@
+<?php declare(strict_types=1);
+
+namespace Surfnet\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20220704084758 extends AbstractMigration
+{
+    public function up(Schema $schema) : void
+    {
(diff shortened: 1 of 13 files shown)


src/Surfnet/Migrations/Version20220704084758.php
```diff
@@ -0,0 +1,28 @@
...
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20220704084758 extends AbstractMigration
+{
+    public function up(Schema $schema) : void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql', 'Migration can only be executed safely on \'mysql\'.');
+
+        $this->addSql('ALTER TABLE vetted_second_factor CHANGE vetting_type vetting_type VARCHAR(255) DEFAULT NULL');
```
pablothedude, line 18: Is it really needed to use NULL? A Self vetting type is still a vetting type, or am I missing something here?

pablothedude (approved):

src/Surfnet/Migrations/Version20220704084758.php
MKodde: This migration is not correct! The UP should add the column, and the down should remove it. This one got recorded incorrectly. Fixing this..+

PR #370, opened 2022-07-13
Support recovery token interactions in RA

MKodde:
Some features where missing to achieve this most notable are:

1. Several data was missing in the recovery tokens projection. For example we need the identity name, email, and institution in  order to inform the RA on which specific Identity owns the RT.
2. The RT DTO did not return the IdentityId
3. Searching RT's was very limited. Several query parameters where added (see docs for new options)

See: https://www.pivotaltracker.com/story/show/181934078
See: https://www.pivotaltracker.com/story/show/181968091

### DIFF
--- a/src/Surfnet/StepupMiddleware/ApiBundle/Doctrine/Type/RecoveryTokenStatusType.php
+++ b/src/Surfnet/StepupMiddleware/ApiBundle/Doctrine/Type/RecoveryTokenStatusType.php
@@ -1,0 +1,92 @@
+<?php
+
+/**
+ * Copyright 2022 SURFnet bv
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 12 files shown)


src/Surfnet/StepupMiddleware/ApiBundle/Doctrine/Type/RecoveryTokenStatusType.php
```diff
@@ -0,0 +1,92 @@
...
+        throw new ConversionException(sprintf("Encountered inconvertible second factor status '%s'", (string) $value));
+    }
+
+    /**
+     * @param mixed $value
+     * @throws ConversionException
+     */
+    public function convertToPHPValue($value, AbstractPlatform $platform): RecoveryTokenStatus
+    {
+        if ($value === '0') {
```
pablothedude, line 71: Seems a bit counter intuitive to use 0 as active...
You could consider using 'active'/'revoked'/'forgotten' instead of '0'/'10'/'20'.
But feel free to leave it as-is.

pablothedude (approved):

src/Surfnet/StepupMiddleware/ApiBundle/Doctrine/Type/RecoveryTokenStatusType.php
MKodde: Good feedback, I opted to follow the existing solution. Which is beneficial from a consistency point of view.

PR #373, opened 2022-07-28
Support vetting type hints (creation and retrieval)

MKodde:
Add support for vetting type hits. A new command was added that allows creation of the vetting type hints for an institution. This command is invoked from RA.

Reading the results of this action is performed by both RA and SS.

See: https://www.pivotaltracker.com/story/show/182129438


pablothedude (approved):
A Command test is missing or should be configured. See unit tests.

PR #375, opened 2022-08-25
Respond with OK when identity not found

MKodde:
Previously we returned a failed response when an identity is not found
in Middleware. This behavior changes as per [1]. We now return an
OK-with-an-empty-array response.

1: https://www.pivotaltracker.com/n/projects/1163646/stories/182988799


quartje:
This is on par with the other apps. I've tested it, and it provides the wanted output.

PR #377, opened 2022-09-28
Wrap address and name into an Address

MKodde:
The to and from methods of the `TemplatedEmail` ask for addresses as its parameter. They are resolved quite magically. Setting a string with a plain mail address is possible, adding multiple parameters is cool too.

The issue we ran into was that the addres param was followed by a common name string. Both where resoved as being mail addresses. Causing the application to blow up on the common name. The solution is to provide Address objects. An address consists of an address and a name.

See: https://www.pivotaltracker.com/n/projects/1163646/stories/183362518
See: https://symfony.com/doc/4.4/mailer.html#email-addresses

### DIFF
--- a/src/Surfnet/StepupMiddleware/CommandHandlingBundle/Identity/Service/RecoveryTokenMailService.php
+++ b/src/Surfnet/StepupMiddleware/CommandHandlingBundle/Identity/Service/RecoveryTokenMailService.php
@@ -133,7 +133,7 @@ final class RecoveryTokenMailService
         ];
 
         $message = (new TemplatedEmail())
-            ->from($this->sender->getEmail(), $this->sender->getName())
+            ->from(new Address($this->sender->getEmail(), $this->sender->getName()))
             ->to(new Address($email->getEmail(), $commonName->getCommonName()))
             ->subject($subject)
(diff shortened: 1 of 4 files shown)


MKodde:
@phavekes I have not tested the bugfix in my environment. But I'm quite certain this resolves the issue(s). Can you verify this on test2?

phavekes (approved):

tvdijen:
I guess indeed nobody tested this before `5.0.1` was merged....

```
Sep 29 17:35:35 sv2010531 STEPUP-MIDDLEWARE[7676]: {"channel":"app","level":"CRITICAL","message":"Call to a member function getEmail() on string","context":{"exception":{"class":"Error","message":"Call to a member function getEmail() on string","code":0,"file":"/apps/installation/Stepup-Middleware/Stepup-Middleware-5.0.1/src/Surfnet/StepupMiddleware/CommandHandlingBundle/Identity/Service/EmailVerificationMailService.php:140"}},"extra":{"server":"middleware.stepup-tst.ssc-ict.overheid-i.nl","application":"middleware","request_id":"18c6ff9d398eae1c965d3fd07a52f7df"}}
```

PR #383, opened 2022-10-25
Bugfix: Call to a member function getEmail() on string

tvdijen:


MKodde (approved):
Thanks for cleaning up after me. 

This neighborhood of the code is hard to test (as demonstrated by the host of bugfix tickets lately..) Our behat test coverage does not hit all mail messages, as painfully demonstrated.

PR #388, opened 2022-11-03
Create the sso_on_2fa feature toggle

MKodde:
In order to enable/disable the SSO on 2FA feature, a feature toggle per institution is required. Allowing us to enable the feature on a per institution basis.

The amount of work required to achieve this task is vast to say the least. In short:

1. The existing `NewInstitutionConfigurationCreatedEvent` was updated with the new option. And a new `SsoOn2faOptionChangedEvent` was created.
2. Support for the new `SsoOn2faOption` was added to the `Command` an `CommandHandler`.
3. Database storage (in the `insittution_configuration` projection) was realized.
4. Tests where updated.

See: https://www.pivotaltracker.com/story/show/183402518


src/Surfnet/StepupMiddleware/GatewayBundle/Entity/InstitutionConfiguration.php
```diff
@@ -0,0 +1,47 @@
+<?php
+
```
Shaky212, line 2: strict types?

Shaky212 (approved):
Looks good in my opinion. But difficult for me to review and give meaningful feedback at this stage. Maybe another general thing to check the strict_types with the new files.

src/Surfnet/StepupMiddleware/GatewayBundle/Entity/InstitutionConfiguration.php
MKodde: That would be awesome, but as a rule of thumb, we only state this on new, or heavily refactored files. That way you can be sure every execution path is tested and verified to be  type strict/safe.

MKodde: Good call, fixed this for all new (non test) classes.

PR #389, opened 2022-11-03
Fix misleading institution configuration error message

pmeulen:
As discussed in https://www.pivotaltracker.com/n/projects/1163646/stories/181933607/comments/234093784


MKodde (approved):
Looks like a great change :+1: 

Maybe target this change's branch to the `feature/sso-2fa` branch?

PR #392, opened 2022-12-08
Improve verification mail message timing

MKodde:
The previous solution where the RegistrationEmailProcessor would send the registration mail messages was decomissioned. As this would also send reg mail messages for other vetting types (self asserted, self vetting).

Now, the SS SP will trigger a SendSecondFactorRegistrationEmailCommand via the middleware client bundle. And that command is now sending the mail message via the existing RegistrationMailService.

Some of the control logic that was in the processor mas moved to the mail service.

See: https://www.pivotaltracker.com/story/show/183826628
See: https://github.com/OpenConext/Stepup-Middleware-clientbundle/pull/96
See: https://github.com/OpenConext/Stepup-SelfService/pull/276


src/Surfnet/StepupMiddleware/CommandHandlingBundle/Identity/Service/RegistrationMailService.php
```diff
@@ -79,12 +122,75 @@ public function __construct(
...
+        $identity = $this->identityService->find($identityId);
+        $institution = new Institution($identity->institution->getInstitution());
+        $institutionConfigurationOptions = $this->institutionConfigurationOptionsService
+            ->findInstitutionConfigurationOptionsFor($institution);
+        $verifiedSecondFactor = $this->secondFactorService->findVerified(new SecondFactorId($secondFactorId));
+
+        if ($institutionConfigurationOptions->useRaLocationsOption->isEnabled()) {
+            $this->logger->notice('Sending a registration mail with ra locations');
+
+            $this->sendRegistrationEmailWithRaLocations(
```
Shaky212, line 145: Perhaps it was a conscious choice but methods sendRegistrationEmailWithRas and sendRegistrationEmailWithRaLocations contain a lot of duplicate code and actually differ in 1 parameter. Perhaps the code could be made simpler and more readable.

Shaky212 (approved):
Looks good. Just a comment on the readability of a piece of code. But you can leave it as it is.

PR #395, opened 2023-01-23
Add identity_vetted column to Second Factor entity

MKodde:
As per point 1 of https://www.pivotaltracker.com/story/show/184191063


MKodde:
This PR was merged with #396, no PRE SAT release is needed to ensure a robust rolling upgrade 'path'.

PR #396, opened 2023-01-24
Update SAT projection

MKodde:
The final 4.5 release already added the second_factor.vetting_type column (later renamed to: `identity_vetted`). Adding it here is not possible. 

The Migration name (timestamp) was updated to be next in sequence to the 4.5 migration.

See task 3 of: https://www.pivotaltracker.com/story/show/184191063

### DIFF
--- a/src/Surfnet/StepupMiddleware/GatewayBundle/Entity/SecondFactor.php
+++ b/src/Surfnet/StepupMiddleware/GatewayBundle/Entity/SecondFactor.php
@@ -96,11 +96,16 @@ class SecondFactor
     private $secondFactorIdentifier;
 
     /**
-     * @var string
+     * This boolean indicates if the second factor token was vetted
+     * using one of the vetting types that are considered 'identity-vetted'.
+     * That in turn means if the owner of the second factor token has its
+     * identity vetted (verified) by a RA(A) at the service desk. This trickles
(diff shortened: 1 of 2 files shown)


src/Surfnet/StepupMiddleware/GatewayBundle/Entity/SecondFactor.php
```diff
@@ -96,11 +96,16 @@ class SecondFactor
...
-     * @var string
+     * This boolean indicates if the second factor token was vetted
+     * using one of the vetting types that are considered 'identity-vetted'.
+     * That in turn means if the owner of the second factor token has its
+     * identity vetted (verified) by a RA(A) at the service desk. This trickles
+     * down to the self-vet vetting type. As the token used for self vetting
+     * was RA vetted.
      *
-     * @ORM\Column(length=255)
+     * @ORM\Column(type="boolean", options={"default":"1"})
```
pmeulen, line 106: Is it expected that Doctrine translates this to `VARCHAR(255) NOT NULL DEFAULT \'1\''`?

MKodde: Ahh, that is a remainder of when this column held the `vetting_type` which was a varchar. I'm going to update this. :+1:

Shaky212 (approved):

PR #397, opened 2023-01-31
Filter non identity-vetted RA candidates

MKodde:
You require a LoA 3 with a vetted idenity token in order to be a RA(A) candidate. The Middelware would later verify this action in the Identity aggregate. But showing only available options in the RA application is good practice. So the filtereing logic to find the available RA candidates was adjusted slightly.

See: https://www.pivotaltracker.com/story/show/184306171


Shaky212 (approved):
Thanks for explaining, looks good!

PR #405, opened 2023-05-15
Docker: Create a parameters.yaml.dist that works in a docker dev env

quartje:
This PR is part of the larger project where we aim to build docker containers of all Stepup apps. All the application .dist files are modified so they will, when instantiated, out of the box with the other docker containers (with a docker-compose file that can be found here: https://github.com/OpenConext/OpenConext-devconf/blob/main/stepup/docker-compose.yml

### DIFF
--- a/config/legacy/parameters.yaml.dist
+++ b/config/legacy/parameters.yaml.dist
@@ -58,10 +58,10 @@ parameters:
 
     email_verification_window: 3600 # the amount of seconds the email verification email/url is valid
 
-    stepup_loa_loa1: https://dev.openconext.local/authentication/loa1
-    stepup_loa_loa2: https://dev.openconext.local/authentication/loa2
-    stepup_loa_loa3: https://dev.openconext.local/authentication/loa3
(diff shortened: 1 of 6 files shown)


config/legacy/parameters.yaml.dist
```diff
@@ -46,35 +46,39 @@ parameters:
...
+    selfservice_api_password: sa_secret
+    registration_authority_api_password: ra_secret
+    readonly_api_password: secret
+    management_password: secret
+    lifecycle_password: secret
 
-    self_service_email_verification_url_template: https://selfservice.tld/verify-email?n={nonce}
-    email_sender_name: SURFnet bv
-    email_sender_email: noreply@surfnet.nl
+    self_service_email_verification_url_template: https://sa.dev.openconext.local/verify-email?n={nonce}
```
MKodde, line 55: Shouldn't this be: `selfservice.dev.openconext.local` ?

MKodde (approved):
NIce one! Ship-it!

PR #407, opened 2023-08-15
Create Gateway Institution Configuration projection row if not exists

MKodde:
When the Gateway institution configuration is updated because the sso on 2fa config changed for that institution. The projection would only be updated if the institution already exists in the projection. If the projection did not yet have the institution, then the projection was not updated.

This fix now creates the row if the projection is not yet present. This only works now that we only track one single configuration value. 

Note that the real solution to this problem would be to do an event replay. But to prevent an event replay on `prod` we choose to add this construction.

https://www.pivotaltracker.com/story/show/185833305


johanib (approved):
lgtm

PR #410, opened 2023-09-21
Fix: the /var/cache/prod directory must be writable by the apache php module

pmeulen:


quartje (approved):
As discussed, this is more cosmetic in nature as it does not solve the original problem. It does make it more precise and readable :)

pmeulen:
> As discussed, this is more cosmetic in nature as it does not solve the original problem. It does make it more precise and readable :)

I'm leaning towards using sudo. See https://github.com/OpenConext/OpenConext-devconf/pull/4#issuecomment-1731280037. What do you think?

Anyway, what we have now does solve the issue.

PR #415, opened 2024-01-04
Add input validation for Activation Code and Document Number

tvdijen:
This PR:

- Restricts document numbers to 6 characters, being A-Z, 0-9 or a hyphen
- Restricts activation codes to 8 characters, being A-Z or 0-9
- Replaces test-values with 'real-life' values instead of 'dumb' dummy-values like 'REGCODE'

Debatable:

- The value for an 'unknown' document was a non-default character (I think it's an 'En Dash' from Extended ASCII table No. 150), but I couldn't get the regexp to match it. Since it really doesn't seem to represent anything (seems to be stored as NULL in the database), I've changed the code to accept `null` for unknown documents.

Related: https://github.com/OpenConext/Stepup-RA/pull/310

### DIFF
--- a/src/Surfnet/Stepup/Identity/Entity/VerifiedSecondFactor.php
+++ b/src/Surfnet/Stepup/Identity/Entity/VerifiedSecondFactor.php
@@ -35,6 +35,8 @@ use Surfnet\StepupBundle\Service\SecondFactorTypeService;
 use Surfnet\StepupBundle\Value\SecondFactorType;
 use Surfnet\StepupBundle\Value\VettingType as StepupVettingType;
 
+use function preg_match;
+
 /**
  * A second factor whose possession has been proven by the registrant and the registrant's e-mail address has been
  * verified. The registrant must visit a registration authority next.
@@ -89,10 +91,10 @@ class VerifiedSecondFactor extends AbstractSecondFactor
         SecondFactorType $type,
         SecondFactorIdentifier $secondFactorIdentifier,
         DateTime $registrationRequestedAt,
-        $registrationCode
+        string $registrationCode
     ) {
-        if (!is_string($registrationCode)) {
-            throw InvalidArgumentException::invalidType('string', 'registrationCode', $registrationCode);
+        if (!preg_match('/^[A-Z0-9]{8}$/i', $registrationCode)) {
+            throw InvalidArgumentException::invalidType('valid characters', 'registrationCode', $registrationCode);
         }
 
         $secondFactor = new self;
@@ -123,7 +125,7 @@ class VerifiedSecondFactor extends AbstractSecondFactor
      * @param SecondFactorIdentifier $secondFactorIdentifier
      * @return bool
      */
-    public function hasRegistrationCodeAndIdentifier($registrationCode, SecondFactorIdentifier $secondFactorIdentifier)
+    public function hasRegistrationCodeAndIdentifier(string $registrationCode, SecondFactorIdentifier $secondFactorIdentifier)
     {
         return strcasecmp($registrationCode, $this->registrationCode) === 0
             && $secondFactorIdentifier->equals($this->secondFactorIdentifier);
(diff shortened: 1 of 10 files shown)


src/Surfnet/Stepup/Identity/Entity/VerifiedSecondFactor.php
```diff
@@ -89,10 +91,10 @@ public static function create(
         SecondFactorType $type,
         SecondFactorIdentifier $secondFactorIdentifier,
         DateTime $registrationRequestedAt,
-        $registrationCode
+        string $registrationCode
     ) {
-        if (!is_string($registrationCode)) {
-            throw InvalidArgumentException::invalidType('string', 'registrationCode', $registrationCode);
+        if (!preg_match('/^[A-Z0-9]{8}$/i', $registrationCode)) {
```
MKodde, line 96: I love to have a more specific input validation here. It might be a good idea to introduce a registration code value object that is tasked with validating the code? That's more in line with the way the domain is set up. For now we kept it primitive because we did not realy bother with elaborate validation.

src/Surfnet/Stepup/Identity/Value/DocumentNumber.php
```diff
@@ -21,47 +21,54 @@
...
     private $documentNumber;
 
     /**
      * @return self
      */
-    public static function unknown()
+    public static function unknown(): self
     {
-        return new self('—');
+        return new self(null);
```
MKodde, line 39: Does this not cause problems when an identitie is forgotten (right to be forgotten) and the entry pops up in the audit log? The value that is set for an unknown document number is displayed then and there in the audit log.

src/Surfnet/Stepup/Identity/Value/DocumentNumber.php
```diff
@@ -21,47 +21,54 @@
...
     /**
-     * @param string $documentNumber
+     * @param string|null $documentNumber
      */
-    public function __construct($documentNumber)
+    public function __construct(?string $documentNumber)
     {
-        if (!is_string($documentNumber) || empty($documentNumber)) {
+        if ($documentNumber === null) {
+            // Created using the static ::unknown method
```
MKodde, line 48: Some work remains to be done here, also see my previous comment about a forgotten Identity.

MKodde (commented):
Hi Tim, I'm a bit late with reviewing this code. I'm about to start upgrading MW to SF6.4 so I wanted to clear as many PR's as possible before having at that.

I'm not comfortable at merging your PR at this stage. I/We need to do some thorough testing and impact analysis before we can go ahead with this.

For now I did leave you some code review pointers. And be aware that I'm going to upgrade the MW. This includes removing the develop branch and renaming master -> main. So be aware of that too.

src/Surfnet/Stepup/Identity/Value/DocumentNumber.php
tvdijen: This is really beyond my understanding... If you can get the regexp to work with this 'funny' character, we can leave this 'as-is'...

src/Surfnet/Stepup/Identity/Entity/VerifiedSecondFactor.php
tvdijen: If you want to go the extra mile, then please be my guest. I'm not going to solve the world's problems here ;)

thijskh: In this case I think it's just a first guard on "sane" input as a defense-in-depth: does this even look like a registration code? The actual validity of the given code will be verified later obviously. So I can imagine that it's just fine to do the basic check that Tim implemented.

src/Surfnet/Stepup/Identity/Value/DocumentNumber.php
pmeulen: This is a defense in depth measure. Since we do not know what the format is of the document numbers – these are whatever the issuer decided to put on the document – retroactively setting a limit in the middleware in this way can cause issues when there are existing document numbers in the event stream that do not validate anymore. Limiting what is allowed in is fine, but this should be done on entry. We could do this in the middleware, but as a defense in depth it is best placed in the RA when the form post is received, not in the middleware.

SInce we do generate the activation codes, we know exactly what they look like so we could validate them in the middleware without additional risk.

tvdijen: I did that, but Peter didn't want it there :-/

PR #417, opened 2024-02-19
Use RA roles when testing authorizations

MKodde:
Before these changes, the authorizations handed out to an Identity when logging in to RA could be escalated unjustly.

The FGA authorizations were previously evaluated to calculate the authz for an Identity. Where they (ofcourse) have to be based on the authorizations that are actually accredited to the Identity.

For many authz decisions in the applications, we do evaluate the Institution authorizations, but when it comes to the determination of the roles for an identity we now correctly look to the `ra_listing` projection.

Merge targets: 
- `release/5.1`
- `main`


src/Surfnet/StepupMiddleware/ApiBundle/Authorization/Service/AuthorizationContextService.php
```diff
@@ -69,40 +69,46 @@ public function __construct(
         $this->authorizationRepository = $authorizationRepository;
     }
 
+    public function buildSelectRaaInstitutionAuthorizationContext(IdentityId $actorId): InstitutionAuthorizationContext
```
pablothedude, line 72: :100: for splitting up the contexts

pablothedude (approved):

PR #421, opened 2024-05-23
Migrate security-check to reusable Github Actions

phavekes:
See https://docs.github.com/en/actions/using-workflows/reusing-workflows


quartje (approved):
awesome, merging!

Issue #492, opened 2024-11-30
Deprovisioning a user fails when already deprovisioned


phavekes:
[This issue is imported from pivotal - Originaly created at Sep 19, 2022 by Bart Geesink](https://www.pivotaltracker.com/story/show/183312425)

When I try to deprovision a user from Middleware which has been deprovisioned before, it will fail with the following message:

{"name":"OpenConext Stepup","status":"FAILED","data":[],"message":"Operation on this Identity is not allowed: it has been forgotten"}

That should not be an error, the deprovision command should be idempotent. Other apps will answer an OK, with empty data:
[{"name":"EngineBlock","status":"OK","data":[]}

I suggest we do the same for middleware

Issue #465, opened 2024-11-29
Middleware deprovision does not deprovision user information


phavekes:
After calling the deprovisioning API

`curl -X "DELETE" -u lifecycle:password -H 'Accept: application/json' "https://middleware/deprovision/urn:collab:person:institution.nl:peter12345" -s`

The response suggest deprovisioning was handled correctly:

```
[...]

Issue #447, opened 2024-11-29
Middleware deprovisioning API dry-run not working


phavekes:
The middleware api documentation indicated a GET request to the deprovisioning endpoint will start a dry-run for the deprovidioning. When calling the API with the GET method, an error is returned instead.

```
curl -X "GET" -u lifecycle:password -H 'Accept: application/json' "https://middleware/deprovision/urn:collab:person:surf.nl:peter12345" -s

```

```
{"errors":["Symfony\\Component\\HttpKernel\\Exception\\MethodNotAllowedHttpException: No route found for \u0022GET \/deprovision\/urn:collab:person:surf.nl:peter12345\u0022: Method Not Allowed (Allow: DELETE)"]}⏎                                                                    
```

johanib:
Make sure the documentation in middleware is up-to-date.
Also check how EB deprovisions, that spec is leading when in doubt.

phavekes:
Tested, still present in 6.0.0-beta3

PR #521, opened 2025-01-15
Fix missing deprovision data

pablothedude:
https://github.com/OpenConext/Stepup-Middleware/issues/465
https://github.com/OpenConext/Stepup-Middleware/issues/447
https://github.com/OpenConext/Stepup-Middleware/issues/492

Acceptance test in devconf:
https://github.com/OpenConext/OpenConext-devconf/pull/33

### DIFF
--- a/src/Surfnet/StepupMiddleware/ApiBundle/Controller/DeprovisionController.php
+++ b/src/Surfnet/StepupMiddleware/ApiBundle/Controller/DeprovisionController.php
@@ -22,7 +22,6 @@ use Exception;
 use Surfnet\Stepup\Exception\DomainException;
 use Surfnet\Stepup\Helper\UserDataFormatterInterface;
 use Surfnet\StepupMiddleware\ApiBundle\Service\DeprovisionServiceInterface;
-use Surfnet\StepupMiddleware\ApiBundle\Controller\AbstractController;
 use Symfony\Component\HttpFoundation\JsonResponse;
 
 class DeprovisionController extends AbstractController
@@ -42,11 +41,12 @@ class DeprovisionController extends AbstractController
             if ($userData !== []) {
                 $this->deprovisionService->deprovision($collabPersonId);
             }
-        } catch (DomainException) {
+        } catch (DomainException $e) {
             // On domain exceptions, like when the identity is forgotten, we return OK, with empty data
             // just so the deprovision run does not end prematurely. At this point, no other domain exceptions
             // are thrown.
             $userData = [];
+            $errors = [$e->getMessage()];
         } catch (Exception $e) {
             $userData = [];
             $errors = [$e->getMessage()];
@@ -57,8 +57,11 @@ class DeprovisionController extends AbstractController
     public function dryRun(string $collabPersonId): JsonResponse
     {
         $this->denyAccessUnlessGrantedOneOff(['ROLE_DEPROVISION']);
+
         $errors = [];
         try {
+            $this->deprovisionService->assertIsAllowed($collabPersonId);
+
             $userData = $this->deprovisionService->readUserData($collabPersonId);
         } catch (Exception $e) {
             $userData = [];
--- a/src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/RaListingRepository.php
+++ b/src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/RaListingRepository.php
@@ -237,4 +237,9 @@ class RaListingRepository extends ServiceEntityRepository
             ->getQuery()
             ->execute();
     }
+
+    public function contains(IdentityId $identityId): bool
+    {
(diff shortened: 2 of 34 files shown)


src/Surfnet/StepupMiddleware/ApiBundle/Controller/DeprovisionController.php
```diff
@@ -57,6 +67,22 @@ public function deprovision(string $collabPersonId): JsonResponse
     public function dryRun(string $collabPersonId): JsonResponse
     {
         $this->denyAccessUnlessGrantedOneOff(['ROLE_DEPROVISION']);
+
+        $nameId = new NameId($collabPersonId);
+        $identity = $this->identityRepository->findOneByNameId($nameId);
+
+        if ($identity === null) {
+            throw new RuntimeException('Cannot forget an identity that does not exist.');
```
johanib, line 75: Since this is a controller, might be better to throw `HttpInvalidParamException`?

src/Surfnet/StepupMiddleware/ApiBundle/Identity/Repository/RaListingRepository.php
```diff
@@ -237,4 +237,9 @@ public function removeByIdentityIdAndInstitution(IdentityId $identityId, Institu
             ->getQuery()
             ->execute();
     }
+
+    public function contains(IdentityId $identityId): bool
+    {
+        return count(parent::findBy(['identityId' => (string)$identityId])) > 0;
```
johanib, line 243: Interestingly `parent::findBy` is used instead of `$this->findBy` :thinking: 
Do you know if there's a reason? Or just copy/paste?

src/Surfnet/StepupMiddleware/ApiBundle/Controller/DeprovisionController.php
```diff
@@ -30,6 +36,9 @@ class DeprovisionController extends AbstractController
     public function __construct(
         private readonly DeprovisionServiceInterface $deprovisionService,
         private readonly UserDataFormatterInterface $formatHelper,
+        private readonly IdentityRepository $identityRepository,
+        private readonly SraaRepository $sraaRepository,
+        private readonly RaListingRepository $raListingRepository,
```
johanib, line 41: If you allow me to be critital (I'll assume yes :stuck_out_tongue:):
Should the dry run be handled by a service aswel?
Three new services injected smells like the controller does more than it should.

src/Surfnet/StepupMiddleware/CommandHandlingBundle/Tests/Identity/CommandHandler/RightToBeForgottenCommandHandlerTest.php
```diff
@@ -178,6 +175,11 @@ public function an_identity_may_not_be_forgotten_twice(): void
             ->with(new IsEqual($nameId), new IsEqual($institution))
             ->andReturn($this->createIdentity($identityId->getIdentityId()));
 
+        $this->sensitiveDataService
+            ->shouldReceive('forgetSensitiveData')
```
johanib, line 179: I would expect some test to fail now because you actually clear the name / email.

Do you think that should/could be tested?

johanib (commented):
Looks neat! :medal_sports:

src/Surfnet/StepupMiddleware/ApiBundle/Controller/DeprovisionController.php
pablothedude: Moved the logic to the service and will ultimately result in a 200 for a dry-run

src/Surfnet/StepupMiddleware/ApiBundle/Controller/DeprovisionController.php
pablothedude: Moved the logic to the deprovision service itself.  I didn't want to move it to a separate service because the dry-run should also handle the deprovisioning, only without the call to the deprovision method on the aggregate. So when changes are made on the deprovisioning this should also be dealt with for the dry-run. And by doing this in the same service one should be aware of this.

src/Surfnet/StepupMiddleware/CommandHandlingBundle/Tests/Identity/CommandHandler/RightToBeForgottenCommandHandlerTest.php
pablothedude: I've looked into this but projections are not tested, I presume this was a design choice because you should be able to replay a projection. I've checked if I could add some tests while at it, but it is hard to do now because of how the compiler passes are implemented for replays (when using an acceptance test) and the repositories are completely mocked away in the current test setup which would take me a long time to setup which I think wouldn't justify the time that would take at the moment.

johanib (approved):

PR #524, opened 2025-01-22
Update dependencies

pablothedude:
@phavekes be aware that this will also adds a config installer from the monitor bundle into MW. I suspect this is intentional but I still wanted to let you know. It seems this also results in two configs:

`config/packages/open_conext_monitor.yaml`
`config/routes/open_conext_monitor.yaml`

It's possible to disable the above, but  I want you to be aware an installer is added.
https://github.com/OpenConext/Stepup-Middleware/pull/524/files#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34R131

https://github.com/OpenConext/Monitor-bundle/issues/14

### DIFF
--- a/src/Surfnet/StepupMiddleware/ApiBundle/Request/CommandValueResolver.php
+++ b/src/Surfnet/StepupMiddleware/ApiBundle/Request/CommandValueResolver.php
@@ -43,12 +43,19 @@ class CommandValueResolver implements ValueResolverInterface
 
         $this->assertIsValidCommandStructure($data);
 
-        $commandName = [];
-        preg_match('~^(\w+):([\w\\.]+)$~', (string)$data['command']['name'], $commandName);
+        $commandNameParts = [];
+        $found = preg_match('~^(\w+):([\w\\.]+)$~', (string)$data['command']['name'], $commandNameParts);
+        if ($found !== 1) {
+            $message = sprintf('Command does not have a valid command name %s', (string)$data['command']['name']);
+            throw new BadCommandRequestException(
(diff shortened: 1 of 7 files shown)


src/Surfnet/StepupMiddleware/ApiBundle/Request/CommandValueResolver.php
```diff
@@ -43,12 +43,17 @@ public function resolve(Request $request, ArgumentMetadata $argument): iterable
 
         $this->assertIsValidCommandStructure($data);
 
-        $commandName = [];
-        preg_match('~^(\w+):([\w\\.]+)$~', (string)$data['command']['name'], $commandName);
+        $commandNameParts = [];
+        preg_match('~^(\w+):([\w\\.]+)$~', (string)$data['command']['name'], $commandNameParts);
+        if (count($commandNameParts) < 1) {
```
johanib, line 48: Should that not be `< 2` or `< 3` since on lince 55/56 indexes 1 and 2 are used?

pablothedude: But the regex will make sure it's >1 or 0.
I shall add a test and use the result of the `preg_match` to hopefully make it more readable.

johanib: ah, yeah. Because there are 2 captures :+1:

johanib (approved):

PR #550, opened 2025-08-19
Check already registered on prove possession

pablothedude:
We add a check to prevent the same token having multiple in flight registration processes for the same token.

https://github.com/OpenConext/Stepup-Gateway/issues/462
https://github.com/OpenConext/Stepup-SelfService/issues/463


pablothedude:
I've also added this check in the Identity aggregate in the move token logic (which is used by a console command) to comply with this business rule.

PR #566, opened 2025-12-09
Fix: Deprovision controller returns error strings instead of array

### LINKED ISSUE #495: Deprovision fails with user lifecycle
[This issue is imported from pivotal - Originaly created at Aug 15, 2022 by Bart Geesink](https://www.pivotaltracker.com/story/show/182988799)

Two issues are pressing:

1. UserLifecycle does not support multiple error messages per pop. Having an array of error messages is not allowed. This can be easily fixed on the Openconext-user-lifecycle side. `https://github.com/OpenConext/OpenConext-user-lifecycle/pull/62`
2. Stepup-MIddleware is moaning to loudly when an identity is not found. This now results in a FAILED response from StepUp. That\'s a bit to severe. When an identity is not using StepUp, it should simply respond with an `OK` with empty data.

johanib:
Prior to this change, the DeprovisionController would return an array of error messages in case something went wrong. This is not the standard behaviour of stepup components. This change ensures the error message is a string.

User lifecycle handles both an array of strings and a string.

Also add a test to ensure a notice is logged if an unknown user is deprovisioned.

Fixes https://github.com/OpenConext/Stepup-Middleware/issues/495

### DIFF
--- a/src/Surfnet/Stepup/Tests/Helper/UserDataFormatterTest.php
+++ b/src/Surfnet/Stepup/Tests/Helper/UserDataFormatterTest.php
@@ -57,9 +57,7 @@ class UserDataFormatterTest extends TestCase
                 ['name' => 'name1', 'value' => 'some-value-1'],
             ],
             'name' => 'Stepup-Middleware',
-            'message' => [
-                'The application is teetering on the edge of catastrophe!',
-            ],
+            'message' => ['The application is teetering on the edge of catastrophe!'],
         ];
 
         $inputData = [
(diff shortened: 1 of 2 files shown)


MKodde:
Nice work, especially on widening the test coverage!

However.. I do not thing we need to implode the error messages to a comma seperated string. Before I thought that would be the solution to the issue. But the issue was already solved in user lifecycle itself. At least, when I check out `main` of user lifecycle in a container. And run the deprovision console command there. Middleware can return strings or arrays of strings in its error response. And all is logged to the console neatly.

See output below:

```
# ./bin/console deprovision urn:collab:person:kwaak          
Continue with deprovisioning of "urn:collab:person:kwaak"? (y/n) y
The user was removed from 1 service.
See error messages below:

 * OpenConext Middleware: Kwakka, Extra error message 1, Extra error message 9

Full output of the deprovision command:

[{"name":"OpenConext Middleware","status":"FAILED","data":[],"message":"Kwakka, Extra error message 1, Extra error message 9"}]
```

I advise to roll back the implode change, and maybe update the new tests if they are still relevant to the existing behavior.

johanib:
> Nice work, especially on widening the test coverage!
> 
> However.. I do not thing we need to implode the error messages to a comma seperated string. Before I thought that would be the solution to the issue. But the issue was already solved in user lifecycle itself. At least, when I check out `main` of user lifecycle in a container. And run the deprovision console command there. Middleware can return strings or arrays of strings in its error response. And all is logged to the console neatly.
> 
> See output below:
> 
> ```
> # ./bin/console deprovision urn:collab:person:kwaak          
> Continue with deprovisioning of "urn:collab:person:kwaak"? (y/n) y
> The user was removed from 1 service.
> See error messages below:
> 
>  * OpenConext Middleware: Kwakka, Extra error message 1, Extra error message 9
> 
> Full output of the deprovision command:
> 
> [{"name":"OpenConext Middleware","status":"FAILED","data":[],"message":"Kwakka, Extra error message 1, Extra error message 9"}]
> ```
> 
> I advise to roll back the implode change, and maybe update the new tests if they are still relevant to the existing behavior.

:heavy_check_mark: After discussion, makes sense.
This is already fixed in user-lifecycle.

MKodde (approved):
Big fat ✔️

PR #573, opened 2026-01-14
Add GET method to deprovision route

phavekes:
To match other Openconext components, the deprovision API should also accept GET requetst

### DIFF
--- a/src/Surfnet/StepupMiddleware/ApiBundle/Resources/config/routing.yml
+++ b/src/Surfnet/StepupMiddleware/ApiBundle/Resources/config/routing.yml
@@ -43,7 +43,7 @@ deprovision:
   defaults:
     _controller: Surfnet\StepupMiddleware\ApiBundle\Controller\DeprovisionController::deprovision
     _format: json
-  methods: [ DELETE ]
+  methods: [ DELETE, GET ]
 
 identity:
   path: /identity/{id}
--- a/docs/MiddlewareAPI.md
+++ b/docs/MiddlewareAPI.md
@@ -117,8 +117,8 @@ Example of possible error messages. These may differ in the real world, but give
 
 ### Allowed to self-vet using an existing self-asserted token?
 
-Is the Identity allowed to self-vet a token using a self-asserted token (SAT). This is only allowed when the only token(s) 
-possessed by the identity are of the self-asserted token vetting type. If another token type is registered, the 
+Is the Identity allowed to self-vet a token using a self-asserted token (SAT). This is only allowed when the only token(s)
+possessed by the identity are of the self-asserted token vetting type. If another token type is registered, the
 middleware will nudge the user to use that token. As using a SAT token will result in a token with a lowered LoA. As the
 identity of the user is not verified by a RA(A).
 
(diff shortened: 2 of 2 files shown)


pmeulen:
Hmm. The app can both read (HTTP verb GET) and delete (HTTP verb DELETE) to the deprovision endpoint:

https://github.com/OpenConext/OpenConext-user-lifecycle/blob/7ac0445cb17cbb791d4fbd95491d9ecf049eae2c/src/OpenConext/UserLifecycle/Infrastructure/UserLifecycleBundle/Client/DeprovisionClient.php#L93

https://github.com/OpenConext/OpenConext-user-lifecycle/blob/7ac0445cb17cbb791d4fbd95491d9ecf049eae2c/src/OpenConext/UserLifecycle/Infrastructure/UserLifecycleBundle/Client/DeprovisionClient.php#L118

This is what that should do: https://github.com/OpenConext/OpenConext-user-lifecycle/blob/7ac0445cb17cbb791d4fbd95491d9ecf049eae2c/docs/deprovision-information.md#required-api-endpoints

This does not look like a the correct solution because the middleware function that is being called now does not distinguish between GET and DELETE: https://github.com/OpenConext/Stepup-Middleware/blob/7230660d0a64bd1d302bb76b2d259958d9ef0034/src/Surfnet/StepupMiddleware/ApiBundle/Controller/DeprovisionController.php#L35-L54

src/Surfnet/StepupMiddleware/ApiBundle/Resources/config/routing.yml
pmeulen, line 1: See previous comment. It looks like the MW deprovision API does not adhere to the API that lifecycle expects (a GET only returns the user's info, it does not deprovision the user)

pmeulen (changes requested):

src/Surfnet/StepupMiddleware/ApiBundle/Resources/config/routing.yml
pmeulen: @johanib What do you think of this?

johanib: Lifecycle: GET `/deprovision/<collabPersonId>`
 `>` Middleware GET `deprovision/<collabPersonId>` :red_circle:  Does not exist.


DELETE `/deprovision/<collabPersonId>`
 `>` DELETE `deprovision/<collabPersonId>`  :heavy_check_mark: 


DELETE `/deprovision/<collabPersonId>/dry-run`
 `>` DELETE `deprovision/<collabPersonId>/dry-run`  :heavy_check_mark: 
 

Without taking the new GET from Peter into consideration, Pieter is right:
The GET `deprovision/<collabPersonId>` does not exist in middleware. User lifecycle implements in in code & documentation, but it does not work. So, if this is needed, we should add the new GET route to middleware, and return the data.
Usually, GET should not perform mutations, especially in the this context where DELETE is also used.

In summary, I think we should follow the lifecycle specifications. 

@phavekes what's the reason for this PR? Is there another component besides user lifecycle that calls the deprovision endpoints on Middleware?

<details>
  <summary>...</summary>
On the other hand, why is the GET even needed, if the DRY RUN returns the same data? (But then the dry run should just return no data)
</details>

---

Another tweak:
The DELETE `deprovision/<collabPersonId>/dry-run` route in middleware should only accept DELETE requests. It currently accepts all http verbs.

PR #595, opened 2026-05-29
Feature/issue 583 email templates to disk

kayjoosten:

PR #597, opened 2026-06-11
Skip events on identical whitelist and config pushes

### LINKED ISSUE #587: Validate middleware eventstream behaviour
When switching from ansible config pushes to Openconext-Middleware config pushes, the config will be sent way more often. How does this affect the event stream? 

Investigate if a config push to middleware with no changes (pushing the same config as last push) adds events to the eventstream.

kayjoosten:
## Summary

- Add `InstitutionCollection::equals()` for order-independent comparison
- Guard `Whitelist::replaceAll()` — no `WhitelistReplacedEvent` if content unchanged
- Guard `Configuration::update()` — no events if decoded config unchanged
- `InstitutionConfiguration` already had correct equality guards per-field; no change needed

## Why

With manage pushing config hundreds of times per day (per [OpenConext/OpenConext-manage#627](https://github.com/OpenConext/OpenConext-manage/issues/627)), all three replace endpoints (`/management/whitelist/replace`, `/management/configuration`, `/management/institution-configuration`) were adding events on every push regardless of whether anything changed.

## Test plan

- [ ] Unit tests added for whitelist and config idempotent push scenarios
- [ ] Behat tests added in devconf (`mw_idempotent_push.feature`) to verify event stream count does not grow on duplicate pushes

Closes #587

> **Note:** This PR targets `feature/issue-583-email-templates-to-disk` (PR #595) as its base, since it depends on the clean `Configuration::update()` from that branch. It should be rebased onto `main` after #595 merges.

### DIFF
--- a/src/Surfnet/Stepup/Configuration/Configuration.php
+++ b/src/Surfnet/Stepup/Configuration/Configuration.php
@@ -50,6 +50,13 @@ class Configuration extends EventSourcedAggregateRoot implements ConfigurationIn
     {
         $decodedConfiguration = JsonHelper::decode($newConfiguration);
 
+        if ($this->configuration !== null
+            && is_array($decodedConfiguration)
+            && $this->deepKsort($this->configuration) === $this->deepKsort($decodedConfiguration)
+        ) {
+            return;
+        }
+
         $this->apply(
             new ConfigurationUpdatedEvent(
                 self::CONFIGURATION_ID,
@@ -80,5 +87,23 @@ class Configuration extends EventSourcedAggregateRoot implements ConfigurationIn
     public function applyConfigurationUpdatedEvent(ConfigurationUpdatedEvent $event): void
     {
         $this->configuration = $event->newConfiguration;
+    }
+
+    /**
+     * @param array<string|int, mixed> $array
+     * @return array<string|int, mixed>
+     */
+    private function deepKsort(array $array): array
+    {
+        ksort($array);
+        foreach ($array as $key => $value) {
+            if (is_array($value)) {
+                $array[$key] = $this->deepKsort($value);
+            }
+        }
+        if (array_is_list($array)) {
+            usort($array, fn($a, $b) => json_encode($a) <=> json_encode($b));
+        }
+        return $array;
     }
 }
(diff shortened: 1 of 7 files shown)


src/Surfnet/Stepup/Identity/Whitelist.php
```diff
@@ -55,6 +55,12 @@ public static function create(InstitutionCollection $institutionCollection): sel
 
     public function replaceAll(InstitutionCollection $institutionCollection): void
     {
+        assert($this->whitelist !== null, 'replaceAll called before whitelist was initialised');
```
kayjoosten: ? what is this

src/Surfnet/Stepup/Identity/Whitelist.php
```diff
@@ -55,6 +55,10 @@ public static function create(InstitutionCollection $institutionCollection): sel
 
     public function replaceAll(InstitutionCollection $institutionCollection): void
     {
+        if ($this->whitelist->equals($institutionCollection)) {
```
johanib: Pipeline fails: Cannot call method equals() on Surfnet\Stepup\Identity\Collection\InstitutionCollection|null.

src/Surfnet/Stepup/Configuration/Configuration.php
```diff
@@ -81,4 +88,19 @@ public function applyConfigurationUpdatedEvent(ConfigurationUpdatedEvent $event)
...
         $this->configuration = $event->newConfiguration;
     }
+
+    /**
+     * @param array<mixed> $array
+     * @return array<mixed>
+     */
+    private function deepKsort(array $array): array
+    {
+        ksort($array);
```
johanib, line 98: It's unclear here what the keys actually are? Are they ints? If so, is the array already sorted. 

If the actual data types are known/determined, consider adding proper type hints instead of mixed.

johanib: discussed: Should sorting changes trigger new configuration events?

johanib: answer: no

johanib (approved):

